PR #3, opened 2024-03-21
Fix workflows

erikbosch:
Previously no docker containers were published. Also quite some links were dead. Fixed by adding files where missing or removing references to KUKSA Server if that was easier.


.github/workflows/kuksa_databroker_build.yml
```diff
@@ -77,6 +81,8 @@ jobs:
       - name: Generate code coverage
         run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
       - name: Upload coverage to Codecov
+        # To use v4 a token must be specifed by "token: ${{ secrets.CODECOV_TOKEN }}""
```
erikbosch, line 84: We get a warning "Node.js 16 actions are deprecated. " if using v3 (but it works).
If using v4 (to resolve warning) it seems  we need to create a token and upload.
Could possibly be done similar to how we do for other secrets at https://github.com/eclipse-kuksa/.eclipsefdn/blob/main/otterdog/eclipse-kuksa.jsonnet#L29

.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
+# See https://pre-commit.com for more information
```
erikbosch, line 1: We can discuss how useful it is to have pre-commit as we mainly have Rust files, but if we have it we need this file.
If not we should remove the corresponding workflow.

PR #4, opened 2024-04-02
Release prep for 0.4.4

erikbosch:


SebastianSchildt:
Which reminds me, we need t get better updating the cargo lock versions, becasue "tag-wise" there WAS a 0.4.2 release fo databroker as well (it has basically all additons over 0.4.1 making int RISCV compatible. seems we just forgot to changee cargo file. Therefore, my suggestion would be, after merging this and then tagging an actual relase, in main we should immediately bump all cargo versions

SebastianSchildt (approved):
Looking good. I quickly started databroker and databroker-cli containers. Seems to work, but did not do full testing with providers etc,

erikbosch:
> Which reminds me, we need t get better updating the cargo lock versions, becasue "tag-wise" there WAS a 0.4.2 release fo databroker as well (it has basically all additons over 0.4.1 making int RISCV compatible. seems we just forgot to changee cargo file. Therefore, my suggestion would be, after merging this and then tagging an actual relase, in main we should immediately bump all cargo versions

In this PR I update Cargo.lock, so when merged Cargo.lock in main will be updated.
I did not know or did not remember that we triggered a docker build of 0.4.2 (possibly as you triggered the build :-))

image

PR #9, opened 2024-04-08
Fix `--help` in clap

argerus:
This functionality was broken by f38363d4ffc4d50d3ce4d5381112b6c84666e92c, or more specifically with the change to resolver = 2.

The `--help` flag requires a feature in clap to be enabled, which is either a newly introduced requirement, or the feature(s) were somehow included implicitly when using resolver = 1.


erikbosch (approved):
Tested and it works

PR #13, opened 2024-04-11
Delete outdated notice

SebastianSchildt:
NOTICE is outdated and not used anymore

(all required dependencies are seen up-to-date in package manager files, and exact transitive dependencies depend on the build environment, and as such can not be documented in code anyway. For our release artifacts we DO generate sboms, and add them to the artifacts where they belong)

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -1,189 +1,27 @@
+# Notices for Eclipse Kuksa
+
+This content is produced and maintained by the Eclipse Kuksa project.
+
+* Project home: <https://projects.eclipse.org/projects/automotive.kuksa>
+
+## Trademarks
+
+ Eclipse Kuksa is a trademark of the Eclipse Foundation.
+
+## Copyright
+
+All content is the property of the respective authors or their employers. For more information regarding authorship of content, please consult the listed source code repository logs.
+
 ## Declared Project Licenses
 
-This program and the accompanying materials are made available under the terms
-of the Apache License 2.0 which is available at
-<https://www.apache.org/licenses/LICENSE-2.0>
+This program and the accompanying materials are made available under the terms of the  Apache License, Version 2.0
+
+You may obtain a copy of the License at
+
+<http://www.apache.org/licenses/LICENSE-2.0>
 
 SPDX-License-Identifier: Apache-2.0
 
-## Contributors
-* Robert Bosch GmbH - initial API and implementation
-* Microsoft Corporation - initial API and implementation
+## Thirdparty Content
 
-## Third-party Content
-
-| Dependency | Version | License |
-|:-----------|:-------:|--------:|
-|aho-corasick|0.7.18|MIT<br/>Unlicense|
-|ansi_term|0.12.1|MIT|
-|anyhow|1.0.56|MIT OR Apache-2.0|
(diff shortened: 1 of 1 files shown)


erikbosch:
If we are to remove `NOTICE.md` we should also remove it from https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/.github/workflows/create_draft_release.yml#L93, then running that workflow should be no problem. It "entered" the release as the release action I created was based on https://github.com/eclipse/kuksa.val.services/blob/main/.github/workflows/release.yml where we also include NOTICE.md

But I am not sure if we "just" shall delete it. First, many of our license headers as of today refer to the NOTICE file like this:

https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/databroker/src/broker.rs

```
/********************************************************************************
* Copyright (c) 2022 Contributors to the Eclipse Foundation
*
* See the NOTICE file(s) distributed with this work for additional
* information regarding copyright ownership.
*
* This program and the accompanying materials are made available under the
* terms of the Apache License 2.0 which is available at
* http://www.apache.org/licenses/LICENSE-2.0
*
* SPDX-License-Identifier: Apache-2.0
********************************************************************************/
```

If using our style of copyright line it seems to be recommended practice to have a NOTICE file, see https://www.eclipse.org/projects/handbook/#ip-copyright-headers. For kuksa-incubation we have a [NOTICE](https://github.com/eclipse-kuksa/kuksa-incubation/blob/main/NOTICE.md) as well, but without third party info. So should we note have a notice file and include it in the release as long as we refer to it from license headers, as it says "_See the NOTICE file(s) distributed with this work_"

The[ Eclipse recommendations on Notice file](https://www.eclipse.org/projects/handbook/#legaldoc-notice) seem to be quite rigid and possibly outdated if a project auto-generate SBOMs. A pragamatic way if we decide to keep the NOTICE file could possibly be to instead of stating third party licenses explain where to find that info, i.e. explain how/where you can find the data for a specific release, latest main or your specific build.

I think some decision would be good. Either we

NOTICE.md
```diff
@@ -1,189 +1,34 @@
...
-|wasi|0.10.2+wasi-snapshot-preview1|Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT|
-|wasi|0.9.0+wasi-snapshot-preview1|Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT|
-|which|4.2.4|MIT|
-|winapi|0.3.9|Apache 2.0<br/>MIT|
-|winapi-i686-pc-windows-gnu|0.4.0|Apache 2.0<br/>MIT|
-|winapi-x86_64-pc-windows-gnu|0.4.0|Apache 2.0<br/>MIT|
+During build thirdparty components may be included. Databroker is a Rust
+project, all our dependencies are declared in the project's Cargo files.
+Depending on the build date and your build environment different transitive
+dependecies might be included. Artifacts published in this repo contain a list
```
erikbosch, line 32: There are some minor spelling errors on this line and next line that we better shall fix

erikbosch (dismissed):
Some minor spelling errors that we better shall fix before merging, but in general looking good

SebastianSchildt:
Done, still needs a "fresh" approve

PR #16, opened 2024-04-18
Feature: continous subscription

lukasmittag:


lukasmittag:
Okay, I see. I can understand. I wasn't aware of this since the PR of @rafaeling did the same.

rafaeling:
Yes I did the same, I guess I forgot the goal of the PR and the issue context after some time.
When I started working on this I remember to read this here https://github.com/eclipse/kuksa.val/issues/652#issuecomment-1717317496 since signal_type was not an inherent property I started with the approach of adding signal_type to VSS but after Comminity meeting it was decided to use better an optional parameter on the request to "decide implicitly" the type of signal (Change or Continuous).
This PR I think implements a good feature but agree with that it does not really resolve the issue.

proto/kuksa/val/v1/val.proto
```diff
@@ -98,6 +98,7 @@ message SubscribeEntry {
 // Subscribe to changes in datapoints.
 message SubscribeRequest {
   repeated SubscribeEntry entries = 1;
+  optional uint64 interval_ms = 2;
```
BjoernAtBosch, line 101: Findings:
* What is the meaning of "interval_ms"? It should be already clear from the parameter name. Currently, I cannot see (without reading the documentation) if it is minimum, typical, or maximum interval.
* Don't we add detailed documentation about the behaviour in the proto files? I agree with Erik (in the old PR): We should also describe the semantics in the interface definition. Imho, the "outside visible behaviour" always belongs to the contract description of an interface definition.
* In the current definition this parameter is related to the overall subscription, but not to certain signals - which is fine for me if the parameter defines the max update frequency. But the behaviour should be clearly defined if the the subscription comprises signals with different update frequencies.
* What is the behaviour, if signals are updated faster than the max frequency requested by the client: 
  * Does it get "just" some sample value at the moment the update interval is reached,
  * do we send some mean value of the last samples (how would that be calculated?),
  * do we accumulated values and send them as a batch,
  * ...?

If we are unsure about the behaviour to be achieved by this PR, we should rethink merging it ...

BjoernAtBosch (commented):
Just looked at the proto changes. (My Rust knowledge is too poor to rate the rest, currently)

proto/kuksa/val/v1/val.proto
BjoernAtBosch: Sorry, if this comment is not "fully" matching - I initially commented the corresponding PR in the old repo.
I also added a comment in the issue: https://github.com/eclipse/kuksa.val/issues/652

erikbosch:
Agreed in discussion to close this one for now

PR #20, opened 2024-04-24
Disable sdv.databroker.v1 by default

argerus:
- Disable sdv.databroker.v1 by default. It can be enabled by supplying a flag at startup.
- Change databroker-cli to use kuksa.val.v1 by default.

### DIFF
--- a/databroker-cli/src/kuksa_cli.rs
+++ b/databroker-cli/src/kuksa_cli.rs
@@ -36,13 +36,13 @@ const TIMEOUT: Duration = Duration::from_millis(500);
 const CLI_COMMANDS: &[(&str, &str, &str)] = &[
     ("connect", "[URI]", "Connect to server"),
     ("get", "<PATH> [[PATH] ...]", "Get signal value(s)"),
-    ("set", "<PATH> <VALUE>", "Set actuator signal"),
+    ("actuate", "<PATH> <VALUE>", "Set actuator signal"),
     (
         "subscribe",
         "<QUERY>",
         "Subscribe to signals with QUERY, if you use kuksa feature comma separated list",
     ),
-    ("feed", "<PATH> <VALUE>", "Publish signal value"),
+    ("publish", "<PATH> <VALUE>", "Publish signal value"),
     (
         "metadata",
         "[PATTERN]",
@@ -318,7 +318,7 @@ pub async fn kuksa_main(_cli: Cli) -> Result<(), Box<dyn std::error::Error>> {
                                 )?,
                             }
                         }
-                        "set" => {
+                        "actuate" => {
                             interface.add_history_unique(line.clone());
 
                             let (path, value) = cli::split_first_word(args);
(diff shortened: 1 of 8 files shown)


databroker-cli/src/kuksa_cli.rs
```diff
@@ -36,13 +36,13 @@ const TIMEOUT: Duration = Duration::from_millis(500);
...
     ("get", "<PATH> [[PATH] ...]", "Get signal value(s)"),
-    ("set", "<PATH> <VALUE>", "Set actuator signal"),
+    ("actuate", "<PATH> <VALUE>", "Set actuator signal"),
     (
         "subscribe",
         "<QUERY>",
         "Subscribe to signals with QUERY, if you use kuksa feature comma separated list",
     ),
-    ("feed", "<PATH> <VALUE>", "Publish signal value"),
+    ("publish", "<PATH> <VALUE>", "Publish signal value"),
```
lukasmittag, line 45: provide?

databroker-cli/src/sdv_cli.rs
```diff
@@ -42,7 +42,7 @@ const CLI_COMMANDS: &[(&str, &str, &str)] = &[
         "<QUERY>",
         "Subscribe to signals with QUERY, if you use kuksa feature comma separated list",
     ),
-    ("feed", "<PATH> <VALUE>", "Publish signal value"),
+    ("publish", "<PATH> <VALUE>", "Publish signal value"),
```
lukasmittag, line 45: provide?

databroker/src/grpc/server.rs
```diff
@@ -165,23 +174,35 @@ where
         info!("Authorization is not enabled.");
     }
 
-    builder
-        .add_service(
+    let kuksa_val_v1 = {
+        if apis.contains(&Api::KuksaValV1) {
+            Some(kuksa::val::v1::val_server::ValServer::with_interceptor(
```
lukasmittag, line 179: val_server can maybe cause confusion with old kuksa_val_server in c++?

lukasmittag (commented):
Some minor naming discussions

databroker/src/grpc/server.rs
argerus: This was not changed as part of this commit. It is referencing generated code, which having a protobuf service definition like this:

```proto
service VAL {
  ...
}
```

will generate code `<camel case service name>_server::<camel case service name>Server`, i.e. `val_server::ValServer`.

databroker-cli/src/kuksa_cli.rs
argerus: Yes, that crossed my mind as well. I'm not against it, but my idea is that `publish` is that act of publishing a single signal value, while "provide" might be better suited for describing the act of continuously providing a signal. "publish" would then be the logical counterpart to "subscribe".

I suppose we will have a discussion around this in the API design rounds. I just couldn't stand having `feed` still there :smiling_face_with_tear:

lukasmittag (approved):
See the comments above. I agree with them an we can merge this.

PR #21, opened 2024-04-26
Initiate Comprehensive Kuksa Requirements Analysis for Enhanced Version Development

rafaeling:
### This pull request marks the beginning of an in-depth analysis aimed at addressing the various challenges and use cases that contributors and the developer community have encountered with recent versions of Kuksa.
The primary goals of this PR are to:

**1. Analyze Software Requirements:** Review of the existing software requirements against the reported issues and use cases.
 
**2. Lay the Groundwork for Future Updates:** Establish a solid foundation for the development of an updated and improved version of Kuksa.

**3. Identify and Document Issues:** Catalogue the specific difficulties experienced by users of the latest Kuksa versions, creating a clear and structured record of known issues.

### **New Content:**
[Kuksa Analysis](https://github.com/SoftwareDefinedVehicle/kuksa-databroker/blob/docs/kuksa_databroker_api/doc/kuksa_analysis.md)


doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,787 @@
...
+
+8.  As a client I want to connect and interact with Databroker either if it contains data or not.
+
+9.  As a client I want to connect and interact with Databroker either if it contains data or not.
+
+10.  As a client I want to know if there is a provider up and running by getting some sensor or actuator value.
+
+11.  As a client I don’t want to `Subscribe` to target values.
+
+12.  As a client I don’t want to `Subscribe` to attributes values.
```
BjoernAtBosch, line 258: Not yet fully sure about that: What about updating individual components on a running system (without restarting "everything")? In that case certain attributes may change and clients need being notified about that.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,787 @@
...
+
+# Functional requirements
+
+### As Client
+1. As a client I want to have a single service point accessible from my execution environment.
+   
+2. As a client I want to `Get` sensor values or actuator current values of the vehicle and its environment in a uniform manner.
+  - Details:
+    * All sensor and actuator values for which a client have permission to.
+    * Client must provide the path, paths or wildcard.
```
BjoernAtBosch, line 230:
```suggestion
    * Client must provide the paths of the requested signals, which optionally may contain wildcards.
```

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,787 @@
...
+    * I want to receive exactly all the signals’ values I requested without errors.
+  
+3. As a client I want to `Subscribe` to sensor values or actuator current values when some signal value changed -> what about continuous subscription?
+  - Details:
+    * All sensor and actuator values for which a client have permission to.
+    * Client must provide the path, paths or wildcard paths.
+    * I want to receive exactly all the signals’ values I requested without errors.
+
+4. As a client I want to `Subscribe` and be able to apply a filter to receive a sensor value with an interval of x milliseconds only if the value has changed.
+
```
BjoernAtBosch, line 240: Not sure where to place that:
n. A signal can have value out of the set of the defined value restriction/data type but it can also have those additional failure states:
    - "unknown": The signal with the specified path/name is not defined on the system.
    - "access denied": The signal is basically known, but the consumer/provider has no permission to access it.
    - "not provided": The signal is basically known, but currently no provider is serving it or the provider lost the connection to the backend component serving this signal.
    - "invalid": The signal is being served by a provider, but currently has no valid value. This may happen if a specific value cannot be determined because of some (temporary) physical sensor error or a target value is currently not known, e.g. if a window is moved by a passenger via up/down button.

The failure states "unknown" and "access denied" are real error situations, that shall be avoided by a proper system configuration. Nevertheless, they may occur in case of clients requesting "optional signals" and also if users of the system enter some private mode and some signals accessible before may get inaccessible.

The states "not provided" and "invalid" are regular states that may occur during normal operation of the running system and must be properly handled by clients.

BjoernAtBosch (commented):
Adding first bunch of findings

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,787 @@
...
+  - [1. Use Cases](#1-use-cases)
+  - [2. Protobuf API Diagram](#2-protobuf-api-diagram)
+
+# Documentation
+  ### [Terminology](./terminology.md)
+  ### [System Architecture](./system-architecture.md)
+
+# Domain requirements
+## Interface requirements
+### Current Databroker API design
```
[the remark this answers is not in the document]
rafaeling: I will move it to the Appendix section.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,787 @@
...
+
+message GetServerInfoResponse {
+  string name                     = 1;
+  string version                  = 2;
+}
+``` 
+
+# Functional requirements
+
+### As Client
```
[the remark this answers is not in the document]
rafaeling: Suggestion committed.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,787 @@
...
+    * Client must provide the path, paths or wildcard.
+
+6.  As a client I want to `Get` attributes values once.
+
+7.  As a client I want to `Actuate` a target value of an actuator and guarantee that the operation succeeded.
+
+8.  As a client I want to connect and interact with Databroker either if it contains data or not.
+
+9.  As a client I want to connect and interact with Databroker either if it contains data or not.
+
```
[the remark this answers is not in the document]
rafaeling: Suggestion committed.

doc/kuksa_requirementes.md
rafaeling: Suggestion commited.

doc/kuksa_requirementes.md
erikbosch, line 1: Change name of file, remove last `e`

doc/diagrams/Client_subscribes.drawio.png
BjoernAtBosch, line 1: * Same as for "get": I'd prefer error states per data point
* We should also add the "update case": A new value is set by the provider after the `subscribe` call and leads to a update notification.
* The `SetResponse` should be returned independently of the client side: Whether the `set` (by the provider) at first notifies all subscribed clients or if it just triggers a thread to do this and immediately returns, is implementation detail (maybe the set response could be left out here)

doc/diagrams/Provider_subscribes.drawio.png
BjoernAtBosch, line 1: The term "**all** target_values" might suggest that the provider is really subscribing for changing target values of **all** the actuator data points.

I'd prefer "selected" or "desired" target_values.

doc/diagrams/Provider_sets.drawio.png
BjoernAtBosch, line 1: At the `:Databroker` lifeline there could be a "self-pointing" arrow with text "identify data points and store values" (or such)

BjoernAtBosch (commented):
Some comments to the diagrams

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+### As Signal Consumer
+1. As a client I want to have a single service point accessible from my execution environment.
+   
+2. As a client I want to `Get` sensor values or actuator current values of the vehicle and its environment in a uniform manner.
+  - Details:
+    * All sensor and actuator values for which a client have permission to.
+    * Client must provide the paths of the requested signals, which optionally may contain wildcards.
+    * I want to receive exactly all the signals’ values I requested without errors.
+  
+3. As a client I want to `Subscribe` to sensor values or actuator current values when some signal value changed -> what about continuous subscription?
```
erikbosch, line 66: I think we need to separate between at least three use-cases:

- Databroker has not received a new datapoint at all (from CAN, SOME/IP, ...)
- Databroker has received a new Datapoint (with new timestamp), but value is the same
- Databroker has received a new Datapoint (with new timestamp and value)

A typical use could be tire pressure monitoring, you may be interested in getting every observation even if value stays the same, as lack of observations might indicate a problem with the system.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+6.  As a client I want to `Get` attributes values once.
+
+7.  As a client I want to `Actuate` a target value of an actuator and guarantee that the operation succeeded.
+
+8.  As a client I want to connect and interact with Databroker either if it contains data or not.
+
+10.  As a client I want to know if there is a provider up and running by getting some sensor or actuator value.
+
+11.  As a client I don’t want to `Subscribe` to target values.
```
erikbosch, line 87: From a testing perspective would it not be better to state:

"As a client I shall not have the possibility to subscribe to target values"

(if we want that). Then testers can verify that it is not possible. Maybe use "shall"/"shall not" rather than "want"/"don't want" in document

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+What's needed:
+* Introduce a way for clients to tell databroker of their preferred rate of updates.
+* Introduce a way for databroker to tell providers of highest needed frequency of sensor
+  data to which they can then adapt.
+  [probably needs] open stream `databroker` -> `provider`
+
+Other considerations:
+
+Setting the desired rate of update would only make sense for sensors of type CONTINUOUS.
+Sensors of type ON_CHANGE would always provide updates when the value changes.
```
erikbosch, line 408: An interesting longterm topic for ON_CHANGE could be to discuss if we need a possibility to specify a threshold for what is considered as a changed value. Taking tire pressure as example, most likely every reading will be slightly different, but if value change from 2.305 to 2.307 it does not really matter. Setting a threshold of for example 0.2 could be useful if you want to know about significant changes as fast as possible.

doc/kuksa_requirementes.md
erikbosch: Do we need or want to be more clear on how this interval requirement should be interpreted and implemented? 
I can see some different implementations that give partially different behavior

- Poll every x millisecond but send only if value is different compared to the last value sent to client
- Whenever databroker receives a changed value (compared to last value in databroker) check if x millisconds has passed, if so send value to client
- Whenever Databroker receives a value and x milliseconds has passed, compare with previous value sent to client and if changed send to client

Do we want alreayd in requirements to state if implementation shall use a poll-based or event-based mechanism, or do we leave that to the implementation?

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+  - [2. Protobuf API Diagram](#2-protobuf-api-diagram)
+
+# Documentation
+  ### [Terminology](./terminology.md)
+  ### [System Architecture](./system-architecture.md)
+
+# Functional requirements
+
+### As Signal Consumer
+1. As a client I want to have a single service point accessible from my execution environment.
```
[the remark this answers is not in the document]
BjoernAtBosch: @lukasmittag: I did not explicitly wanted to exclude that. The "single service point" was more intended to say that I want to have a "central" (maybe backed up) instance providing access to "all" the signals - instead of having multiple service providing specific signals each (e.g. seat service providing all the seat related signals, wiper service, etc).

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+### As Databroker
+1. The data broker shall be able to handle catalogs of signals described by the syntax as defined by the COVESA Vehicle Signal Specification (VSS). This relates to all aspects of the VSS syntax definition, which is also called VSS rule set. This implies that the data broker can handle the signal catalog as defined by the COVESA VSS.
+
+2. Databroker shall support at least those metadata elements as defined by the VSS rule set.
+
+### As Provider
+1. As a provider I want to `Set` the value and current value of one or many sensors and/or actuators with its timestamp to Databroker.
+2. 
+3. As a provider I want to `Update` using stream atomically the value and current value of one or many sensors and/or actuators with its timestamp to Databroker.
```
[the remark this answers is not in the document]
BjoernAtBosch: * Why distinguishing between _value_ and _current value_? Imho, the value of a sensor could also be labelled as _current value_.
* Adding "efficiently" to emphasize the requirement to do this with least possible overhead
```suggestion
3. A provider shall be able to continuously and efficiently (e.g. via grpc stream) _**update**_ the current value of one or multiple sensors and/or actuators with its timestamp.
```

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+4. As a provider I want to `Subscribe` to actuator target values requests.
```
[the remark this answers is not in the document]
BjoernAtBosch: I also like the idea of having a "mixed document" where the reasoning of why requirements are that way is done "around" the respective requirement and the hard requirements are then clearly visible placed in the text (maybe put in some box or preceded by a well-recognizable bold requirements marker).

Anyway, we later - once we finished this more or less - should add unique ids to the requirements.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+5. As a provider I want to `Set` once attributes values.
```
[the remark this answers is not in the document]
BjoernAtBosch: Yes, attribute values could be part of the signal config (vss.json) passed to the broker at startup.

Question is then: How to update databroker with new signals at runtime (e.g. in case of a "dynamic" update)? Easiest option would be to stop and restart the broker with the new signal config.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+6. As a provider I don’t need to but I would like to `Get` sensor values or current values of actuatros?.
```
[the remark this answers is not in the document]
BjoernAtBosch: I see _consumer_ and _provider_ as separate roles - but not as separate components/applications: A component can have one of those roles or both, i.e. being pure consumer or pure provider, or a combined "consumer provider".
From that point of perspective, this requirement should be removed.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+1.	Any clients of the data broker (applications, providers) shall not assume the availability of the data broker service when they startup.
+2.	Any clients of the data broker (applications, providers) shall not assume the permanent availability of the data broker service during their runtime.
+3.	Any applications/clients shall not assume the availability of a value for any data point at their startup.
+4.	Any applications/clients shall not assume the permanent presence of a value for any data point during their runtime.
+Explanation: Any component of the system can come and go - components could stop working due to a crash (what should not but will happen) or because of an update (which is a regular use case). Furthermore, components could reside on different execution environments which need restarts at different points of time. This shall not result in stopping and restarting the overall system. Instead, each and every component shall react in an appropriate way on changing availability of its dependencies.
```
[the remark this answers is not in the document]
BjoernAtBosch: @lukasmittag: The reason is written in the explanation. You don't know in what order the overall collection of components is started. Also, it can happen that the broker is (temporary) not available - due to crash or update. Same is true for current signal values.
Regarding the state of signals/values:
1. A certain expected signal could be not know/available in a vehicle system, i.e. there is no respective provider installed and executable on the system (consequently, it shall not be part of the "broker config vss.json")
1. The client might have not access (rights) to it
   a. The vehicle manufacturer does not allow the client to access the signal
   b. The user ("driver") disabled access to the signal (like on smart phones: "Is this app allowed to access my camera?")
1. It might be known and accessible, but currently no value is given by a provider
1. It may have an "invalid" state, i.e. the provider is available, but maybe the physical sensor is out of order or has some temporary issues
1. The signal is known, accessible, currently provided, and has a valid value.

* The "failure" states 1 and 2(a) are "permanent". It makes no sense to install a client on the system if it needs _**mandatory**_ request to signals in those states. If the client just uses them _optionally_, it still can make sense.
* The states 3 + 4 are of temporary nature (hopefully) and consumers (and providers) need being able to proper handle these situations.
* The last state is the "ideal case" from the client's point of perspective. But the world isn't ideal ...

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+## Path requests
+The data broker shall support at least those metadata elements as defined by the VSS rule set. Data points/nodes are (primarily) identified ("addressed") by their name/path which is a string.
+VSS arranges nodes in a tree structure, separating elements via a single dot ("."). This shall be supported but must not be a mandatory requirement.
```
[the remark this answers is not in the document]
BjoernAtBosch: @lukasmittag: Yes, I guess the wording is not correct. What's meant is, that the broker shall also support 
* "flat" signal structures (i.e. without any path separator)
* maybe the path separator (the ".") should be configurable.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+```
+
+## Setting Values
+1. Attributes: 
+   * It shall not be possible to set attribute values, except once at startup time by its respective responsible provider.
+2. Sensors: 
+   * There shall be only one client able to set the current sensor value during each arbitrary span of time.
+3. Actuators:
+   * ? Actuator data points have a current and a target value. The current value represents the actual state of the actuator, whereas the target value represents a state desired by that client, who most recently set that target value. 
+   * Only one client shall be able to set the current actuator value during each arbitrary span of time. This client is the provider of the data point.
```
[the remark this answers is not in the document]
BjoernAtBosch: Imho, we should use the [existing terminology](https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/doc/terminology.md) as linked at the beginning of the document, which defines _(VSS) consumers_ and _(VSS) providers_. A _client_ is a provider and/or a consumer.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+   * Multiple client may be able to set the target value of an actuator.
```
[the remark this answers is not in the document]
BjoernAtBosch: We should have a general "requirement", that each signal can have only one provider - at least at a time. Regardless if it are sensors or actuators,
* current values should be set/updated by a single instance (at a time), only,
* only a single instance (at a time) shall react on actuation requests.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+Something conceptually like this:
+1. User calls `set(Vehicle.Door.Locked, field=TARGET_VALUE, true)`
+2. User calls `subscribe(Vehicle.Door.Locked, field=VALUE)`
+3. Provider (subscribing to TARGET_VALUE) receives the request and starts actuating, providing VALUE when it changes.
+4. User is notified when VALUE turns to true, and concludes that the door has now been locked.
+
+But what happens if the door fails to lock, e.g. the door is not close or the actuator is broken?
+* What should the user subscribe to for this information?
+* And how long should it wait before concluding that it failed?
```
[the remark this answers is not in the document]
BjoernAtBosch: The VSS defines an explicit defect sensor signal for some actuators (like the front beams), i.e. `Vehicle.Body.Lights.Beam.Low.IsDefect` of boolean type. But that's not generally the case. 

We could discuss introducing a general defect failure state in addition to the four failure states proposed above. Setting a target value to an actuator in defect state could then be rejected by the broker/provider with a proper error.

But that wouldn't help in situations where the requested target value cannot be "achieved" for other reasons, e.g. seat or window movement is stopped due to a blockage; or a remotely requested seat movement is rejected because the vehicle is currently being driven. Currently, one could just recognize that the seat/window movement is stopped (before reaching the target) or not happening at all via the updates of the current value. (For other actuators, VSS explicitly distinguishes between current and target values). 

I guess we should not try to fix all the shortcomings of the VSS spec ...

BTW: Door's lock and open states are not necessarily related: You could lock an open door and if you close it, you can't reopen it (because of the lock). The actual behaviour will be vehicle specific.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+When designing the databroker API, the method kuksa.val uses for providing signals was investigated. In short, kuksa-val-server takes the approach of trying to shoehorn the need of signal providers into the existing VISSv2 protocol. This is problematic for several reasons:
+
+* By reusing, **but changing the meaning** of the terminology used in VSS, there is a constant dissonance between the terms used. This is a recipe for confusion.
+* By deviating from the standard in _this particular way_, a standards compliant VISSv2 client cannot use it in any meaningful way.
+* It makes it harder to actually provide a standards compliant VISSv2 in the future .
+* By using the same methods for both signal providers _and_ users of signals it's harder (or impossible) to customize them for their different requirements.
+
+With this in mind, databroker chose to make a clear distinction between signal providers and signal users. It doesn't use this terminology though. It does this by splitting the interface into two separate services, which are customized for their different requirements / use cases. It doesn't need to be implemented in this way in order to achieve the same design goal, though.
```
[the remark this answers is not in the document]
rafaeling: I think @argerus is the original author of this section, maybe he can do some rephrasing :)

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,787 @@
...
+
+  rpc GetMetadata(MetadataRequest) returns (MetadataResponse);
+
+  rpc GetServerInfo(GetServerInfoRequest) returns (GetServerInfoResponse);
+}
+
+message EntryRequest {
+  string path                     = 1;
+  View view                       = 2;
+  repeated Field fields           = 3;
```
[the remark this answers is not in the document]
rafaeling: True

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+  - [Registration of Datapoints](#registration-of-datapoints)
+  - [Availability of Datapoints](#availability-of-datapoints)
+  - [Lifecycle of components](#lifecycle-of-components)
+  - [Path requests](#path-requests)
+  - [Errors](#errors)
+  - [Setting Values](#setting-values)
+  - [Atomic operatoins](#atomic-operatoins)
+  - [Update notifications](#update-notifications)
+  - [Access rights](#access-rights)
+  - [VSS signals - Users vs providers](#vss-signals---users-vs-providers)
```
eriksven, line 26: To me the whole document more reads like a mix between requirements, design considerations, design decisions and and API definition. Maybe it becomes more accessible and when we split the document into different documents which can be discussed separately. Otherwise, I think it will be very different and lengthy to achieve consensus on the whole PR. 

After a first read, this the part before the discussion of the `VSS signals - Users vs providers` could be in another document than what follows afterwards.

doc/kuksa_requirementes.md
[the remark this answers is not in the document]
eriksven: Do you mean "single service endpoint" here? Just after reading your comments I understand that "single service point" here is not a datapoint in the databroker (which still left the question what a service point is compared to a regular datapoint :|)

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+  - [1. Use Cases](#1-use-cases)
+  - [2. Protobuf API Diagram](#2-protobuf-api-diagram)
+
+# Documentation
+  ### [Terminology](./terminology.md)
+  ### [System Architecture](./system-architecture.md)
+
+# Functional requirements
+
+### As Signal Consumer
```
eriksven, line 57: I get a bit confused with the usage of the term `signal consumer` and `client`. 

The Terminology defines that a client is either a consumer or a provider. The headline seems to limit the scope to clients which are consumers. However, the following statements start with `a client` and not with `a signal consumer` is this by intention? Looking at the subsequent paragraph for providers, my feeling is that it should rather be changed to `a signal consumer`.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+1. As a client I want to have a single service point accessible from my execution environment.
+   
+2. As a client I want to `Get` sensor values or actuator current values of the vehicle and its environment in a uniform manner.
+  - Details:
+    * All sensor and actuator values for which a client has permission to.
```
eriksven, line 62: `actuator current`

The formulation was a bit misleading for me at first. How about `client can only get values for signals to which it has permission to`
I first thought, that a single `Get` should return all available values which I guess is not the intention.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+# Documentation
+  ### [Terminology](./terminology.md)
+  ### [System Architecture](./system-architecture.md)
+
+# Functional requirements
+
+### As Signal Consumer
+1. As a client I want to have a single service point accessible from my execution environment.
+   
+2. As a client I want to `Get` sensor values or actuator current values of the vehicle and its environment in a uniform manner.
```
eriksven, line 60: how about `of the vehicle signals` instead of `vehicle and its environment` because I am not sure how to re-define vehicle and its environment when we already have definition for signal.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+  - Details:
+    * All sensor and actuator values for which a client has permission to.
+    * Client must provide the paths of the requested signals, which optionally may contain wildcards.
```
eriksven, line 63: Do we really want to allow wildcards for `Get`? This is discussed below too. To my understanding potentially processing wildcards in the Get function may add processing overhead/latency to the commonly used `Get` endpoint. 
Because of that I would only allow wildcards in the `GetMetadata` which is not called that often and has less performance requirements. A typically flow could be to call `GetMetadata` once and then use the result in `Get`.

doc/kuksa_requirementes.md
BjoernAtBosch: My idea was/is to have a document containing both design considerations together with the resulting requirements. The requirements could be tagged, so that they can be filtered from the rest later.

doc/kuksa_requirementes.md
BjoernAtBosch: In my initial document this was part of a "basic requirement". Signals of the vehicle's environment would be outside air temperature, for example

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+    * I want to receive exactly all the signals’ values I requested without errors.
```
eriksven, line 64: Can we clarify this to something like: 
`I want to receive exactly all the signal values I requested or an error if retrieving at least on the requested signals fails` 
(not sure whether this is the current behavior)

doc/kuksa_requirementes.md
eriksven: In addition to what @erikbosch writes, I would add to the sentence at least something like `to sensor values or actuator current values AND GET THEN NOTIFIED when some signal value changed`.

eriksven: For me there is also room for interpretation whether a client can subscribe to multiple signals and is then notified  with all signals when one signal changes or only with the changed signals. 

For instance, the client subscribes with a single call to A, B, and C. After B changes, does the client get only B or also A and C where the values remain the same. My intuition would say only B, but just reading the requirement both seems possible to me.

doc/kuksa_requirementes.md
BjoernAtBosch: I don't think, that a consumer should rely on all the requested signals can be received without "errors". I wrote about states of signal (values) above. Some state might represent error situations, some just might be temporary lacking a valid value. I do **not** believe, that we can release consumers from handling those kind of situations (maybe there are some very simple consumers/app, where this would be possible. But more mature consumers will need - and want - to handle that failures.) 
Furthermore, if requesting multiple signals in one request. The failures shall be reported per signal separately.

doc/kuksa_requirementes.md
erikbosch: Does it necessarily need to add that much overhead?, I mean if you first make a lookup on exact match, and only do wildcard/regxp matching if that fails.

doc/kuksa_requirementes.md
eriksven: I See, and linking the design considerations and the requirements sounds good. Given the overall length of the document, I would still opt to keep it on multiple pages and use cross-references. 
In the long run, I would hope that the documents that we get out of this process can be the foundation for creating something like a Kuksa API specification in a similar style as other specs like VISSv2 (https://github.com/COVESA/vehicle-information-service-specification) 

Maybe more of a philosophical question, but should the requirements be more the result of the use cases and user expectations, and then the design considerations be the answer on how to address the requirements?

doc/kuksa_requirementes.md
eriksven: Then I interpreted the `environment` too wide. I was more thinking about the information about traffic lights, charging infrastructure, attached trailers etc.

doc/kuksa_requirementes.md
eriksven: Good point, that might work. I was more thinking from the perspective of performance and promising maximum processing delays in the future (when running on specific hardware). For these cases, we would then clearly need to communicate that the performance varies depending on whether wildcards are used or not.

doc/kuksa_requirementes.md
eriksven: Answering to the comment by @erikbosch 

Is the requirement `with an interval of x miliseconds` a new requirement (feature request)? I was not aware that the current version of the databroker supports this. 

Intuitively, to me subscribe still has the semantic of only getting informed about changed values. Because of that,  I would go for the third option to check what has been sent to the client before and potentially adding "grace period" before after the last value has been sent before sending another value.

But I get that there are uses cases as well where the client is interested in the value at a fixed frequency to essentially downsample the data stream coming from the sensor.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+4. As a client I want to `Subscribe` and be able to apply a filter to receive a sensor value with an interval of x milliseconds only if the value has changed.
+   
+5.  As a client I want to `GetMetadata` static Metadata from signals.
+  - Details:
+    * All sensors, actuators, and attributes values for which a client has permission to.
+    * Client must provide the path, paths or wildcard.
+
+6.  As a client I want to `Get` attributes values once.
```
eriksven, line 79: Why the `once`?

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+7.  As a client I want to `Actuate` a target value of an actuator. This value has to be forwarded to the provider of this actuator if the provider is available, otherwise the target value is lost.
```
eriksven, line 81: Just to be sure, this is a massive change to the current behavior, right? My expectation in the current databroker would have been that the target value simply stays in the databroker no matter whether someone consumed the data or not. 

Now, it seems that you only want to forward target values if there is already someone subscribed to this value who can perform the actuation (aka actuation provider). This requires a 1:1 connection between the target of a signal and an already subscribed provider. 

I am a bit afraid that we limit ourselves too much here. For instance, how about cases where a third-party is interested in what actuation if currently going on (or failed)? I could also think about cases, where the target value is still valid even there was no provider in the moment it was set. Here the provider could check the timestamp and decide for itself whether the actuation shall be performed or not.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+5.  As a client I want to `GetMetadata` static Metadata from signals.
+  - Details:
+    * All sensors, actuators, and attributes values for which a client has permission to.
+    * Client must provide the path, paths or wildcard.
+
+6.  As a client I want to `Get` attributes values once.
+
+7.  As a client I want to `Actuate` a target value of an actuator. This value has to be forwarded to the provider of this actuator if the provider is available, otherwise the target value is lost. 
+
+8.  As a client I want to connect and interact with Databroker either if it contains data or not.
```
eriksven, line 83: What the reason for the `either if it contains data or not`. Are there other cases that I am missing which are implicitly excluded by the requirement?

doc/kuksa_requirementes.md
BjoernAtBosch: That's a thing which I can still not fully grab in my head. I was thinking about having a few "basic requirements" what consumers (and providers) can expect from the broker. I.e. things like "I want to pull/get the current state of vehicle signals via a single service interface (in an efficient way)". This does not yet describe certain details. 

Based on that set of requirements some "design considerations" need to be discussed and decided, resulting in a more detailed "next level" set of requirements addressed to towards the API design and broker implementation.

doc/kuksa_requirementes.md
eriksven: I am not sure either whether we need this explicit exclusion. However, the VSS documentation (https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/attributes/) states: 

`Attribute values can also change, similar to sensor values. The latter can be useful for attribute values that are likely to change during the lifetime of the vehicle. However, attribute values should typically not change more than once per ignition cycle, or else it should be defined as a sensor instead.`

This definition would match with the assumption that attributes (should) do not change during an ignition cycle. But do clients "survive" an ignition cycle?

doc/kuksa_requirementes.md
BjoernAtBosch: I was discussing with Rafael and John about specifying a lifetime of a value (also for sensor and current actuator values). This could be also used for target values: If the provider is not "picking" a target value up within its lifetime, it is dropped. That means, that a short "outage" of a provider does not lead automatically to a lost actuation request, but target values set "some hours" before also don't lead to no longer expect actuation. 

John's counter-arg was that the info about temporary missing provider would be given back to the client (via signal specific failure) and then it would be the clients task to retry once the provider "is back" (and the actuation still is wanted).

BjoernAtBosch: Forgot to state, that I think John's point is valid. The "pretended convenience" of storing the target value in the broker for a certain time is a false one: The consumer must be able to handle outages of the broker itself, anyway. Therefore, it's best to implement the retry on consumer side. This could be a convenience functionality offered by the consumer SDK/library.

eriksven: The "pretended convenience" is a good point. 
It still means that the consumer who wants to trigger the actuation needs to find the moment in time when the actuation provider is connected. 
With defining a lifetime of how long the actuation target is valid when setting the target value, one could ease this is a bit. But I am not sure whether we will see many useful actuation provider which really come and leave that frequently so that this becomes relevant.


Would this mean that the databroker only brokers sensor values and for actuation more acts as a router and translator?

doc/kuksa_requirementes.md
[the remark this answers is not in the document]
eriksven: I agree that regular providers do not handle attribute values. If there is a requirement to change attributes during runtime, this might be an argument for this attribute actually being a sensor. 

We had a number of concerns with changing the VSS tree during runtime (adding/removing nodes etc.), so I would changes to attributes only through the signal config at the start (as proposed by @BjoernAtBosch )

doc/kuksa_requirementes.md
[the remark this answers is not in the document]
eriksven: The notion of defining different roles for consumers and providers while both perspectives can be implemented in the same component makes sense to me. 
I would however, still state what each roles requires to do to make explicit what the roles mean. 

So for the specific case, I think it makes sense for a provider get the current values at least for the sensors that it provides since it might need to know the indicated value in the databroker.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+2. 
+3. As a provider I want to `Update` using stream atomically the value and current value of one or many sensors and/or actuators with its timestamp to Databroker.
+
+4. As a provider I want to `Subscribe` to actuator target values requests.
+
+5. As a provider I want to `Set` once attributes values.
+
+6. As a provider I don’t need to but I would like to `Get` sensor values or current values of actuatros?.
+   
+7. As a provider I don’t want to `Set` target values to Databroker.
```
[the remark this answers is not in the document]
eriksven: Agreed, unless there is a hard reason to now allow something. In this case, I agree we do not explicitly exclude the provider from being able set target values.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+## Data availability/persistence according to lifecycle of Client, Databroker and Provider
+1. The data broker either ...
+   * Stores last set values during its own lifecycle,
+   * ? stores values during the system's power cycle (i.e., "persists" values over own restarts, or
+   * store values over system's power cycles.
+
+2. How to "reset" values availability if its provider got inactive (without reseting the value)?
+   * -> Client's job (e.g. using timestamp)?
+   * -> Broker's job (e.g. using timestamp + minimal update cycle)?
```
eriksven, line 125: I do not have clear opinion, but am torn towards leaving the decision whether a value is still considered as available to the client. This might depend on the circumstances, the use case and the signal and it is therefore hard to make a general decision even per signal. 

However, it would of course be great if the broker could already decide whether the signal is considered as not available anymore, to relive the client from making this evaluation and avoid sending around the data and timestamps. 
What do you think about allowing the client to optionally add a minimal update cycle to the request and then letting the databroker decide per request whether the signal is returned? This would of course add processing load to the databroker.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+3. Provider and client aliveness
+   * If there is no active client subscription should the provider stop sending values to Databroker?
+   * If there is no active provider setting values while client subscription? Should Databroker or Client be aware of it?
```
eriksven, line 129: Would be nice to be aware of whether there is an active provider.  But depending on the signal and time since the last update to the signal, I would still allow the consumer to get the data even if there is not an active provider in that moment.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+   * Stores last set values during its own lifecycle,
+   * ? stores values during the system's power cycle (i.e., "persists" values over own restarts, or
+   * store values over system's power cycles.
+
+2. How to "reset" values availability if its provider got inactive (without reseting the value)?
+   * -> Client's job (e.g. using timestamp)?
+   * -> Broker's job (e.g. using timestamp + minimal update cycle)?
+
+3. Provider and client aliveness
+   * If there is no active client subscription should the provider stop sending values to Databroker?
```
eriksven, line 128: I think it should still send values. There is still a high chance that a client is requesting single values through Get and without subscribing.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+   * If there is no active provider setting values while client subscription? Should Databroker or Client be aware of it?
+
+## Wildcard support
+* Should `Get` and `Subscribe` calls support Wildcard request?
+* Do we want it only for `GetMetadata`?
```
eriksven, line 133: See discussion above. Maybe remove this questions here in favor of what we come up above.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+Reference -> [Wildcard](./wildcard_matching.md)
+
+## Registration of Datapoints
+Do we need a method for providers to register data points at runtime? **Implemented in old API?**:
```
eriksven, line 138: I think we should not have this option. In the old API it had a number of shortcoming also in terms of security and potential denial of service and I would keep the signal tree as static as possible.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+
+Its purpose would be:
+1.	Publishing the availability of new data points during runtime
```
eriksven, line 141: What kind of data point do you expect to show up during runtime which are not known at design time/configuration time of the databroker? 

To me there is a difference between signals which are currently not available or that will not be available at all. With the VSS tree I would express which signals can become available in that broker even if they might not be available at the moment.

doc/kuksa_requirementes.md
argerus: > Would this mean that the databroker only brokers sensor values and for actuation more acts as a router and translator?

Yes. Or put another way:
- Signal values (data) are cached.
- Actuation requests (requests, possibly expecting a response) are not cached.

Even if we decide that actuation requests should not return a response, it is very hard / impossible to formulate a consistent cache invalidation policy for actuation requests if we were to model them as "setting a field in a database", i.e. "target values".

It does not play well with actuators that can be controlled not only through databroker - from other ECUS over CAN or directly connected physical buttons - which I expect would be a pretty standard situation. When that happens, the target value is outdated and pretty much useless.

At the time of writing, I don't know a single provider implementation that checks the timestamp of a target value or at any point "clears" them. Neither are any consumers that I'm aware of. They essentially live forever. I think it's a lot easier to write robust providers (and consumers) by modelling actuation as a request / response thing.

doc/kuksa_requirementes.md
```diff
@@ -0,0 +1,805 @@
...
+2.	Claiming providership of those data points
+
+
+In a mature system some "central instance", e.g., the system integrator must be responsible to ensure that no two components can register/claim providership for any data point.
```
eriksven, line 145: How would one ensure that there is no contention at design time? To me it sounds like the databroker should still manage and check such a "ownership". One way for a provider could be to open a stream which writes to the signal. The broker then would only allow a single stream at the same time with write access to a current value.

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+## 1. Kuksa client subscribes on changed sensor values or actuator current values of all signals
+
+**Primary Actor:** Kuksa client
+
+**Secondary Actors:** Databroker, Provider, CAN 
+
+**Priority**: High
+
+**Preconditions:**
+ * Provider open CAN Bus connection and start receiving signal values.
```
eriksven, line 25: What do you think about abstracting this as "vehicle bus" or "vehicle network" instead of CAN?

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+ * Provider sets sensor values or actuator current values to Databroker atomically. -> NFR (1.7)
```
eriksven, line 26: Not sure where NFR 1.7 links to. Is this Non Functional Requirements from the kuksa_requirements.md ? But I did not find an entry for 1.7 there.

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+![Client Subscribes](./diagrams/Client_subscribes.drawio.png)
+
+**Basic Flow:**
+1.	Kuksa client subscribe to all the sensor values or current values of actuators.
+2.	Kuksa client start receiving streaming of data.
+3.	Use case finishes.
+
+**Alternative Flows**:
+* Flow 1.
+2.	Databroker return error response since provider is not connected???
```
eriksven, line 51: Do we only allow the stream to be opened once there is already a provider available? Or can we have mechanism that there is currently not a provider available and leave it to the consumer whether to wait for a provider to start providing or to close the stream and try it later.

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+## 2. Kuksa client sets a target value for an actuator.
+
+**Primary Actor:** Kuksa client
+
+**Secondary Actors:** Databroker, Provider, CAN Bus
+
+**Priority:** High
+
+**Preconditions:**
+* Provider is connected to some CAN Bus.
```
eriksven, line 66: How about: `provider is able to trigger execution of change in actuator`
There are various ways of how an actuation provider might realize a set target value, where sending a CAN message is just one option.

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+* Provider started sending signals current values to Databroker before the use case starts.
```
eriksven, line 67: Why is sending the current value before the actuation happens a precondition here?

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Special requirements:**
+
+**Assumptions:**
+
+**Postconditions:**
+* Client receives a response whose status indicates the operation was successfully executed.
+* Databroker stores on his database the new target value for the actuator.
+* Provider writes the new target value in CAN bus.
+
```
eriksven, line 79: To me, there is a difference between the operation being successfully executed and writing a target value to the bus. The post condition sounds like the client already receives the success message even the actuation is not even finished yet. 
The underlying question is: What do we consider as semantics for a successful actuation?

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+**Sequence diagram:**
+![Client Actuate](./diagrams/Client_actuate.png)
+
+**Basic Flow:**
+1.	Kuksa client calls actuate with a valid actuator path and keeps waiting for a response.
+2.	Databroker checks if a Provider is connected.
+3.	Databroker changes its internal target value for the corresponding actuator path with the current timestamp. (Assumption -> Current value remains the same)
+4.	Databoker notifies all subscriptions to target values.
```
eriksven, line 87: Question: Do we allow multiple subscribers to the target value of which only one is registered as the provider?

doc/kuksa_use_cases.md
```diff
@@ -0,0 +1,214 @@
...
+5.	Provider subscribes to target values on Databroker.
```
eriksven, line 88: This sounds more like a pre-condition and part of the flow.

doc/kuksa_use_cases.md
eriksven: This applies to the other use cases too.

doc/kuksa_requirementes.md
rafaeling: I will rephrase all the requirements to support this naming.

doc/kuksa_requirementes.md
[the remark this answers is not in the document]
rafaeling: Also agree, will remove negated requirements.

doc/kuksa_requirementes.md
rafaeling: I think we aligned on that some time ago and understood that the decision was to return only a grpc error with detailed information on failed signals when dealing with a singles request to multiple signals as specified here (https://github.com/SoftwareDefinedVehicle/kuksa-databroker/blob/docs/kuksa_databroker_api/doc/kuksa_requirementes.md#errors).
So for me the requirement would be as you said @erikbosch:
`I want to receive exactly all the signal values I requested or an error if retrieving at least on the requested signals fails`

doc/kuksa_requirementes.md
rafaeling: I will modify the requirements for support `signal consumer` terminology.

doc/kuksa_requirementes.md
rafaeling: Thanks for the suggestion, will use it.

doc/kuksa_requirementes.md
BjoernAtBosch: You could think of a scenario, where you add new applications and providers via an app store. With that, new signals could become available during runtime. This could be the case if providing hardware is present but software is missing (during delivery of the vehicle). Or "higher value" signals (based on pure software) could be offered this way. User then can download by payment. 

We could handle this via updating the local broker config and then restarting the broker (loading the new config).

eriksven: At least in the hardware example, I would expect that the tree already contains the signals that the providing hardware could provide since the hardware is then known at design time of the system and hardware. 

That was at least my assumption to differentiate between signals that are currently not available (hardware present but provider missing) and that will not be available at all. 

In general, to have the option to add (or remove?) signals during runtime seems to add more complexity and edge cases which is another reason why I like the approach of updating the local broker config more.

erikbosch:
Some thought based on the requirement discussion today - do we for the requirements and use-cases in this PR need to add a "status field" that shows the approval/implementation-status so that it is clear if the requirement/use-case represents a long term goal or actually is implemented and verified.

eriksven:
> Some thought based on the requirement discussion today - do we for the requirements and use-cases in this PR need to add a "status field" that shows the approval/implementation-status so that it is clear if the requirement/use-case represents a long term goal or actually is implemented and verified.

That is very good point, and I agree that tracking the status seems to be needed.

My initial thought was that the initial requirements document would reflect the current state of the implementation and we could then use subsequent PRs to discuss changes and additions which would just be merged once we reached an agreement on the requirement and an initial implementation fullfilling this requirement. But we already moved far away from that approach so we definitly need an approach to track the status/consensus of the different requirements. I am also happy for any solution to break down this PR into smaller fractions which can be merged independently :)

BjoernAtBosch:
> > Some thought based on the requirement discussion today - do we for the requirements and use-cases in this PR need to add a "status field" that shows the approval/implementation-status so that it is clear if the requirement/use-case represents a long term goal or actually is implemented and verified.
> 
> That is very good point, and I agree that tracking the status seems to be needed.
> 
> My initial thought was that the initial requirements document would reflect the current state of the implementation and we could then use subsequent PRs to discuss changes and additions which would just be merged once we reached an agreement on the requirement and an initial implementation fullfilling this requirement. But we already moved far away from that approach so we definitly need an approach to track the status/consensus of the different requirements. I am also happy for any solution to break down this PR into smaller fractions which can be merged independently :)

I am thinking to define each requirement (and maybe each use case) as a Github issue. These could then be linked from other - implementation issues, release notes, and s on). But as Github issues are not well "sortable" we would still need some other document where we could order and link the requirements.

doc/kuksa_requirementes.md
BjoernAtBosch: I disagree putting signal definitions into the config just because the providing hardware is present. I'd prefer to have only that signals in the config, that have a installed (and runnable) provider in the system. "Currently not available" means for me that the respective provider is currently not up and running (i.e. not yet started or temporary inactive due to a crash or update), or that some other component down the chain to the hardware is temporary not active.

doc/kuksa_analysis/kuksa_use_cases.md
```diff
@@ -0,0 +1,243 @@
...
+**Assumptions:**
+ * Signal Consumer subscribes to all the possible sensor values or actuator current values to test special requirements.
+ * Provider porvides signals values to Databroker as fast as receives data from the Vehicle Network.
+
+**Postconditions:**
+ * A subscription is created, and Signal Consumer receives streaming of signals.
+
+**Sequence diagram:**
+
+![Signal Consumer Subscribes](../diagrams/Signal_Consumer_subscribes.png)
```
BjoernAtBosch, line 53: My intention for removing some API-specific details is to have the use cases independent of the implementation.
```suggestion
The following sequence shows a situation, where a consumer subscribes to a set of signals supplied by two different providers. One has already connected to the Databroker when the consumer raises its subscription, the other provider is connecting afterwards.

![Signal Consumer Subscribes](../diagrams/Signal_Consumer_subscribes.png)
```

doc/kuksa_analysis/kuksa_use_cases.md
```diff
@@ -0,0 +1,243 @@
...
+## UC1
+    Title: Signal consumber subscribes to all signal values.
+
+    Status: 🟡
+
+    Description: Signal Consumer subscribes on changed sensor values or actuator current values of all signals
+
+**Primary Actor:** Signal Consumer
+
+**Secondary Actors:** Databroker, Provider, Vehicle Network
```
BjoernAtBosch, line 33: The Databroker isn't an actor - it is passive and reacts to requests of consumers and providers. As all the use cases are related to it, imho we do not need to mention it here at all:
```suggestion
**Secondary Actors:** Provider, Vehicle Network 
```

doc/kuksa_analysis/kuksa_use_cases.md
```diff
@@ -0,0 +1,243 @@
...
+4.	Signal Consumer closes subscription to Databroker.
+    4.1 The Databroker sends a stop signal command with the previous uuid.
+    4.2 The provider stops reading signal values from the Vehicle Network.
+5.	Use case finishes.
+
+**Exceptions:**
+
+
+## UC2
+    Title: Signal Consumer actuates on an actuator.
```
BjoernAtBosch, line 78:
```suggestion
    Title: The consumer wants to set the value of an actuator signal.
```

doc/kuksa_analysis/kuksa_use_cases.md
```diff
@@ -0,0 +1,243 @@
...
+    Description: Signal Consumer subscribes on changed sensor values or actuator current values of all signals
+
+**Primary Actor:** Signal Consumer
+
+**Secondary Actors:** Databroker, Provider, Vehicle Network 
+
+**Priority**: High
+
+**Preconditions:**
+ * Signal Consumer contains a valid authentication token to connect and perform calls to databroker.
```
BjoernAtBosch, line 38:
```suggestion
 * The consumer passes a valid authentication token to authorize having read access rights to all requested signals.
```
or is it: "... token authorizing to have read access ..."?

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+
+# Functional requirements
+
+## As Signal Consumer
+### FR1-ASC
+    Title: Single Access Service Point for Signal Consumer
+
+    Status: 🟡
+
+    Description: The Signal Consumer shall have a single service point accessible from my execution environment.
```
lukasmittag, line 44:
```suggestion
    Description: The Signal Consumer shall have a single service point accessible.
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+   
+### FR2-ASC
+    Title: Uniform Retrieval of Vehicle and Environmental Signal Values
```
lukasmittag, line 47:
```suggestion
    Title: Uniform retrieval of vehicle and environmental signal values
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+
+    Status: 🟡
+
+    Description: The Signal Consumer shall get sensor values or actuator values of the vehicle signals and its environment(air temperature, etc) in a uniform manner.
```
lukasmittag, line 51:
```suggestion
    Description: The Signal Consumer shall get sensor or actuator values of the vehicle signals and its environment(air temperature etc) in a uniform manner.
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+      1.  IF there is an error
+          * THEN the Signal Consumer shall not receive any signal value, just one single error with detailed error information
+          * ELSE the Signal Consumer shall receive exactly all the signals’ states(failure or value) requested.
+
+          n. A signal can have value out of the set of the defined value restriction/data type but it can also have those additional failure states:
```
lukasmittag, line 56:
```suggestion
          n. A signal can have a value in range of the defined restrictions/data type. Additionally it can hold these failure states:
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+            - "unknown": The signal with the specified path/name is not defined on the system.
```
lukasmittag, line 57: Review these states not sure they're correct anymore

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+            - "access denied": The signal is basically known, but the consumer/provider has no permission to access it.
```
lukasmittag, line 58:
```suggestion
            - "access denied": The signal is known, but the consumer/provider has no permission to access it.
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+            - "not provided": The signal is basically known, but currently no provider is serving it or the provider lost the connection to the backend component serving this signal.
```
lukasmittag, line 59:
```suggestion
            - "not provided": The signal is known, but currently no provider is serving it or the provider lost the connection.
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+            - "invalid": The signal is being served by a provider, but currently has no valid value. This may happen if a specific value cannot be determined because of some (temporary) physical sensor error or a target value is currently not known, e.g. if a window is moved by a passenger via up/down button.
+       
+      2.  The Signal Consumer only shall get values for signal to which it has permission to.
+      3.  The Signal Consumer shall provide the paths of the requested signals, which optionally may contain wildcards.
+  
+### FR3-ASC
+    Title: Subscription and Notification for Signal Value Changes
```
lukasmittag, line 66:
```suggestion
    Title: Subscription and Notification for signal value changes
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+
+    Status: 🟡
+
+    Description: The Signal Consumer shall subscribe to sensor values or actuator current values of a signal and then get notified when some signal value changed.
```
lukasmittag, line 70:
```suggestion
    Description: The Signal Consumer shall be able subscribe to sensor or actuator current values of a signal. When subscribed the signal consumer must get notified when there's a change in the signal value.
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+      *  IF there is an error
+          THEN the Signal Consumer shall not receive any signal value, just one single error with detailed error information
+          ELSE the Signal Consumer shall receive exactly all the signals’ states(failure or value) requested.
+      *  The Signal Consumer only shall get values for signal to which it has permission to.
+      *  The Signal Consumer shall provide the paths of the requested signals.
+
+### FR4-ASC
+    Title: Filtered Interval Subscription for Signal Values
```
lukasmittag, line 78: why everything upper case?

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+
+    Status: 🟡
```
lukasmittag, line 80: shouldn't this be red?

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+      *  The Signal Consumer shall provide the paths of the requested signals.
+
+### FR5-ASC
+    Title: Accessing Static Metadata of Signals
+
+    Status: 🟡
+
+    Description: A Signal Consumer shall be able to get static Metadata from signals.
+    - Details:
+      * All sensors, actuators, and attributes values for which a Signal Consumer has permission to.
```
lukasmittag, line 96: Have we agreed on this? I thought metadata is not in the scope of permissions?

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+
+### FR6-ASC
+    Title: Retrieving Attribute Values for Signal Consumption
+
+    Status: 🟡
+
+    Description: A Signal Consumer shall be able to get attributes values.
+
+### FR7-ASC
+    Title: Actuation of Actuator Value with Databroker Confirmation
```
lukasmittag, line 107: Databroker confirmation I would put into brackets () since we will not implement this in a first step

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+
+    Status: 🟡
+
+    Description: The Signal Consumer shall be able to actuate a value of an actuator. This value has to be forwarded to the provider of this actuator if the provider is available, otherwise the target value is lost:
```
lukasmittag, line 111:
```suggestion
    Description: The Signal Consumer shall be able to request an actuation. This request has to be forwarded to the provider of this actuator if the provider is available, otherwise the request gets lost:
```

doc/kuksa_analysis/kuksa_requirementss.md
```diff
@@ -0,0 +1,202 @@
...
+
+    Status: 🟡
+
+    Description: The data broker shall offer a method to providers to update the current state of a set of signals.
+      a. The current state consists a timestamp and either of valid value or a failure state.
+      b. The data broker shall offer a method optimized for frequent updates.
+      c. The data broker should offer a second method for non-frequent updates that is easy to use in a provider's implementation.
+      d. The data broker shall reject updating the current state of signals where the client is not the provider of.
+      e. The data broker shall store the updated value or failure state of a signal.
+
```
lukasmittag, line 197: We could add here one more point regarding VISS. How much we will support it. If we want to support it etc. Maybe also stating the problems we see with VISS

erikbosch:
To be included in 0.5
@rafaeling - check if anything needs to be fixed before merge/approval

doc/kuksa_analysis/kuksa_design_topics.md
```diff
@@ -0,0 +1,744 @@
...
+    SensorCommand sensor_command = 2;
+    Error error = 3;
+  }
+}
+
+...
+
+```
+
+The full `.proto` specification for this proposal can be found [here](https://github.com/boschglobal/kuksa.val/blob/feature/provider_api/kuksa_databroker/proto/sdv/databroker/v1/provider.proto).
```
erikbosch, line 622: This link is dead and caused CI error. In general search for all `boschglobal` and `kuksa.val`, none of them shall be used.

lukasmittag (commented):
should we have an app use case here as well wer we subscribe values then publish values from a provider and finally actuate upon a value from databroker and do the actuation and value update. Maybe not the right scope here just a thought.

README.md
```diff
@@ -114,10 +114,13 @@ The quickest possible way to get Kuksa Databroker up and running.
...
    > :bulb: **Tip:** You can stop the container using `ctrl-c`.
 
+*Note that not all APIs are enabled by default, see [user guide](doc/user_guide.md) and*
+*[protocols](doc/protocol/README.md) for more information!*
+
 ### Reading and writing VSS data using the CLI
 
 1. Start the CLI in a container attached to the _kuksa_ bridge network and connect to the Databroker container:
-   The databroker supports both of `sdv.databroker.v1` and `kuksa.val.v1` as an API. Per default the databroker-cli uses the `sdv.databroker.v1` interface. To change it use `--protocol` option when starting. Chosse eihter one of `kuksa-val-v1` and `sdv-databroker-v1`.
+   The databroker supports both of `sdv.databroker.v1` and `kuksa.val.v1` as an API. Per default the databroker-cli uses the `sdv.databroker.v1` interface. To change it use `--protocol` option when starting. Choose either one of `kuksa-val-v1` and `sdv-databroker-v1`.
```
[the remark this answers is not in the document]
erikbosch: I think this one is inherited from a (cherry-picked?) change I did on main, but I agree that we should update this section before release.

doc/kuksa_analysis.md
```diff
@@ -0,0 +1,16 @@
+# KUKSA Analysis
```
erikbosch, line 1: Shall we have a link to this file from the main README?

doc/kuksa_analysis.md
```diff
@@ -0,0 +1,16 @@
...
+
+This documentation provides a comprehensive analysis of the KUKSA project, detailing essential aspects to ensure a clear understanding and effective implementation. The sections covered include:
+
+### [1. Requirements](./kuksa_analysis/kuksa_requirements.md)
```
erikbosch, line 5: Should we possibly remove the link from the heading, and instead on a line below write something like 

`The requirements for Kuksa Databroker is available in the <kuksa requirement document>`

Right now it is not obvious that the header actually is a link, same for the other sections below

doc/kuksa_analysis.md
```diff
@@ -0,0 +1,16 @@
...
+
+### [2. Design Topics](./kuksa_analysis/kuksa_design_topics.md)
+
+### [3. Use Cases](./kuksa_analysis/kuksa_use_cases.md)
+
+### [4. API Definition](../proto/kuksa/val/v2/val.proto)
```
erikbosch, line 11: This is just one of the APIs we support. If we add text like proposed above we could write that this is our future/present API, the others (which we also could link to) are considered legacy APIs.

doc/kuksa_analysis/kuksa_design_topics.md
```diff
@@ -0,0 +1,725 @@
...
+- [Split subscribe method due to performance reasons](#split-subscribe-method-due-to-performance-reasons)
+- [Service VAL better naming](#service-val-better-naming)
+- [Extend and split service definition for V3?](#extend-and-split-service-definition-for-v3)
+- [COVESA topics](#covesa-topics)
+
+# Data availability/persistence according to lifecycle of Client, Databroker and Provider
+    Status: 🔴
+    Current decisions:
+      1. Databroker stores last signal values during its own lifecycle.
+      2. It is not possible to reset values.
```
erikbosch, line 54: You can send in "None" now and that is practically reset, or?

doc/kuksa_analysis/kuksa_design_topics.md
```diff
@@ -0,0 +1,725 @@
...
+      3. Signal Consumer and Provider are not aware of each other.
+
+    Description:
+   1. The data broker either ...
+      * Stores last set values during its own lifecycle,
+      * ? stores values during the system's power cycle (i.e., "persists" values over own restarts, or
```
erikbosch, line 60: `?`

doc/kuksa_analysis/kuksa_design_topics.md
```diff
@@ -0,0 +1,725 @@
+# Design Topics
```
erikbosch, line 1: We have some open topics in this document, like questions and statements started with `? `

Shall we possibly add a short into to this document saying that this document tries to collect current design decisions as well as topics where we not yet have made any decision.

doc/kuksa_analysis/kuksa_design_topics.md
```diff
@@ -0,0 +1,725 @@
...
+   3. Provider and client aliveness
+      * If there is no active client subscription should the provider stop sending values to Databroker?
+      * If there is no active provider setting values while client subscription? Should Databroker or Client be aware of it?
+
+# Wildcard support
+    Status: 🟢
+    Current decisions:
+      Only `ListMetadata` support wildcard due to complex error handling in the implementation and usability.
+    Description:
+ * Should `Get` and `Subscribe` calls support Wildcard request?
```
erikbosch, line 76: Remove this line? Possibly also mention performance as a reason.

doc/kuksa_analysis/kuksa_design_topics.md
```diff
@@ -0,0 +1,725 @@
...
+  map<int32, Datapoint> entries = 1;
+  //map<int32, Notification> entries = 1; // add to discussion PR
+}
+
+# Service VAL better naming
+    Status: 🔴
+    Current decisions:
+    Description:
+
+# Extend and split service definition for V3?
```
erikbosch, line 713: V3 is V2, or?

Possibly explain what we mean with V3?

doc/pictures/api.svg
erikbosch, line 1: Is this file used anywhere?

rafaeling:
> should we have an app use case here as well wer we subscribe values then publish values from a provider and finally actuate upon a value from databroker and do the actuation and value update. Maybe not the right scope here just a thought.

@lukasmittag those use cases you mention are all covered individually, I thought the same but maybe the use case get so big that is more difficult to understand than just splitting it into small use cases? just an opinion.

erikbosch (approved):
Good enough, we can improve things in separate PRs if needed

PR #24, opened 2024-05-05
CycloneDX SBOM in artifacts

SebastianSchildt:
Ported and improved version of  from https://github.com/eclipse/kuksa.val/pull/756

## Main change in generated artifacts:
Creates and includes  CycloneDX Software Bill of Materials (SBOM) for the databroker. This helps with compliance using kuksa in a commercial context, and is generally also one of the accepted formats fulfilling EU CRA SBOM requirements

## "Collateral improvements"
 - Unified the way databroker and cli is buit locally with Github action builds while retaining workflow parallelism (added parallel integration tests)
 - Using the canonical way to determine dash dependencies for Rust
 - Split the SBOM and license generation, now a CycloneDX SBOM input is required to collect license files. This step has been split out to a Python package in kuksa-common: https://github.com/eclipse-kuksa/kuksa-common/tree/main/sbom-tools , so it can later be reused for non-rust components as well
 - get rid of the createbom python code in this repo -> less code, better


.github/workflows/kuksa_databroker_build.yml
```diff
@@ -301,28 +272,22 @@ jobs:
...
-         python3 createbom.py --dash ${{github.workspace}}/dash-databroker ../databroker
+          cargo tree -e normal --prefix none --no-dedupe -p databroker --target all --all-features  > ${{github.workspace}}/cargodeps
+          cat ${{github.workspace}}/cargodeps | sort -u  \
+          | grep -v '^[[:space:]]*$' | grep -v kuksa | grep -v databroker  \
+          | sed -E 's|([^ ]+) v([^ ]+).*|crate/cratesio/-/\1/\2|' \
+          > ${{github.workspace}}/dash-databroker-deps
+
       - name: Dash license check
-        uses: eclipse-kuksa/kuksa-actions/check-dash@3
+        uses: eclipse-kuksa/kuksa-actions/check-dash@2
```
erikbosch, line 291: Why do you go back to v2? I guess that will cause problems if you run https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/.github/workflows/create_draft_release.yml (as then the Dash artifact has the same name for both databroker and databroker-cli which gives a conflict)

.github/workflows/kuksa_databroker-cli_build.yml
```diff
@@ -207,15 +218,24 @@ jobs:
...
-         python3 createbom.py --dash ${{github.workspace}}/dash-databroker-cli ../databroker
+          cargo tree -e normal --prefix none --no-dedupe -p databroker-cli --target all --all-features  > ${{github.workspace}}/cargodeps
+          cat ${{github.workspace}}/cargodeps | sort -u  \
+          | grep -v '^[[:space:]]*$' | grep -v kuksa | grep -v databroker  \
+          | sed -E 's|([^ ]+) v([^ ]+).*|crate/cratesio/-/\1/\2|' \
+          > ${{github.workspace}}/dash-databroker-deps
 
       - name: Dash license check
-        uses: eclipse-kuksa/kuksa-actions/check-dash@3
+        uses: eclipse-kuksa/kuksa-actions/check-dash@2
```
erikbosch, line 239: Why back to v2?

.github/workflows/kuksa_databroker_build.yml
```diff
@@ -126,26 +115,28 @@ jobs:
       working-directory: ${{github.workspace}}/
       run: |
         which cross || cargo install cross
+        cargo install cargo-license cargo-cyclonedx
+        pip install "git+https://github.com/eclipse-kuksa/kuksa-common.git@6f3d7627760582d8ba83cc8a0f7449d00fffee84#subdirectory=sbom-tools"
```
erikbosch, line 119: Would it be a nice to long term ambition to publish "kuksa-sbom-tools" as pypi package ?

.github/workflows/kuksa_databroker_build.yml
```diff
@@ -126,26 +115,28 @@ jobs:
...
     - name: Build
       working-directory: ${{github.workspace}}/
+      env:
+        KUKSA_DATABROKER_FEATURES: databroker/viss,databroker/tls
```
erikbosch, line 123: How does features works for the CLI? Right now we do not set the same features for CLI - intentional? or is CLI independent from the features, i.e. support VISS/Websocket and TLS anyway?

Dockerfile-cli
```diff
@@ -12,20 +12,16 @@
 # ********************************************************************************/
 
 # This is expected to be executed in the kuksa.val top-level directory
-# You need to run build-all-targets-cli.sh first, as this docker file jsut
+# You need to run build-databroker-cli.sh first, as this docker file just
```
erikbosch, line 15: Maybe include needed arguments, like `./build-databroker-cli.sh amd64 arm64 riscv64`

build-databroker-cli.sh
```diff
@@ -0,0 +1,154 @@
...
+#
+# where platform can be one or more of
+#
+# arm64, amd64, riscv64, i.e. the following are valid commandlines
+#
+# ./build-databroker-cli.sh amd64
+# ./build-databroker-cli.sh amd64 arm64 riscv64
+#
+# you can enable features that will be passed to cargo
+# by setting the environment variable KUKSA_DATABROKERCLI_FEATURES, i.e.
```
erikbosch, line 31: We miss the example feature line:

`# KUKSA_DATABROKERCLI_FEATURES=databroker/viss,databroker/tls`

(But do we need/use any features for CLI? See other comment for features used in CI, we use VISS only for databroker, not databroker-cli, right?)

.github/workflows/kuksa_databroker_build.yml
SebastianSchildt: yes. Or at least tag it. But for now, it being an internal tools that might still change, I think using hash is ok

build-databroker.sh
```diff
@@ -0,0 +1,160 @@
...
+# ./build-databroker.sh amd64
+# ./build-databroker.sh amd64 arm64 riscv64
+#
+# you can enable features that will be passed to cargo
+# by setting the environment variable KUKSA_DATABROKER_FEATURES, i.e.
+#
+# KUKSA_DATABROKER_FEATURES=databroker/viss,databroker/tls
+#
+# If you want generate an SBOM and assemble a list of licenses set
+# KUKSA_DATABROKER_SBOM to "yes" or "true",
```
erikbosch, line 37: `y`as well?

build-databroker-cli.sh
SebastianSchildt: I did not add an example, becasue those I think are not valid CLI features, and I dit not know, what were

licensecuration.yaml
```diff
@@ -0,0 +1,22 @@
...
+  # OpenSSL seems to be in the process to move to Apache-2.0, but this
+  # may take some time
+  ring:
+    expression: "ring AND ISC AND OPENSSL"
+
+
+expressions:
+    # LLVM exception not relevant for the project, thus in situation
+    # like this we choose to use vanilla Apache-2.0 terms
+    "Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT": "Apache-2.0"
```
erikbosch, line 22: @mikehaller - we discussed license curation some time ago, could be interesting to check if this curation file is considered by our internal scanning tools.

erikbosch:
General comments:

- The Dockerfiles contains some references to  `build-all-targets*`, replace them with info on how to use the new files if you want to build locally
- Shall we possibly give some info on how to build Docker containers locally in README?

I believe this change will break the https://github.com/boschglobal/kuksa-databroker/blob/feature/cyclonedx-sbom/.github/workflows/create_draft_release.yml. Hopefully not if we use latest kuksa-action dash version. Anyway, should better be tested.

Adding @lukasmittag and @argerus as reviewers, so that they are aware of the upcoming change and can protest, if needed

build-databroker-cli.sh
erikbosch: Then just remove the `, i.e.`

.github/workflows/kuksa_databroker_build.yml
SebastianSchildt: copy & paste fail. Should be fixed

Dockerfile-cli
SebastianSchildt: added

build-databroker-cli.sh
SebastianSchildt: done

build-databroker.sh
SebastianSchildt: 👍

SebastianSchildt:
I think I addressed all comments, I just did not add "how to build docker locally" documentation. 
As it was not there before, do you think it is required? The Dockerfile itself is not "magic", and there is a hint, you need to run the build script first inline

licensecuration.yaml
SebastianSchildt: Hmm I stand corrected. It CAN express general rules. Anyway, as you see there were a lot more "package specific" curations needed for some reason. But anyway, different topic maybe :D

.github/workflows/kuksa_databroker-cli_build.yml
```diff
@@ -207,15 +218,24 @@ jobs:
...
+          cat ${{github.workspace}}/cargodeps | sort -u  \
+          | grep -v '^[[:space:]]*$' | grep -v kuksa | grep -v databroker  \
+          | sed -E 's|([^ ]+) v([^ ]+).*|crate/cratesio/-/\1/\2|' \
+          > ${{github.workspace}}/dash-databroker-deps
 
       - name: Dash license check
         uses: eclipse-kuksa/kuksa-actions/check-dash@3
         with:
-          dashinput: ${{github.workspace}}/dash-databroker-cli
+          dashinput: ${{github.workspace}}/dash-databroker-deps
```
erikbosch, line 241: The name (`dash-databroker-deps`) must be different than the name used for databroker, otherwise aggregated builds (like create draft release) will end up  with an error:

`Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run`


... as then both jobs use the same upload area.

If you change here (and on line 236) to `dash-databroker-cli-deps` then there should be no name conflict.

Test build at https://github.com/boschglobal/kuksa-databroker/actions/runs/8965789017

SebastianSchildt: Fixed

build-databroker.sh
```diff
@@ -0,0 +1,160 @@
...
+        echo "$(ls ${TARGET_MAP})"
+        rm -rf ${TARGET_MAP}
+        exit 1
+    fi
+done
+
+
+# Starting a fresh build
+echo "Cargo clean, to start fresh..."
+cargo clean
```
argerus, line 149: This breaks caching.

I don't think it's a good default for a script to force removal of all build caches. That is better controlled outside the script or added as an optional argument if deemed useful. Since each platform is built in a separate runner, we do not need to do a cargo clean in CI anyway (that is why it was added, right?). For one thing, it removes the cached install of `cargo cross`, which makes the build slower. It also removes the cache of dependencies, further making the build slower.

argerus (changes requested):

.github/workflows/kuksa_databroker-cli_build.yml
```diff
@@ -80,19 +85,56 @@ jobs:
...
         which cross || cargo install cross
-    - name: Building
+        cargo install cargo-license cargo-cyclonedx
+        pip install "git+https://github.com/eclipse-kuksa/kuksa-common.git@6f3d7627760582d8ba83cc8a0f7449d00fffee84#subdirectory=sbom-tools"
+    - name: Build
       working-directory: ${{github.workspace}}/
+      env:
+        KUKSA_DATABROKERCLI_SBOM: y
+      run: |
+        ./build-databroker-cli.sh ${{ matrix.platform.name }}
```
argerus, line 100: I think it would be a lot better if we avoid reintroducing SBOM generation / license checking to the build step. All the other "sanity checks" are run in parallel, i.e. unit tests / formatting etc. The artifacts (SBOM) are not needed until after the build has completed anyway (for the release artifact creation).

SebastianSchildt: Yes, but it would be a bit harder to refactor and I guess not save as much time overall, as you still need to do it 3 times (the sbom is different, depending on the target, i.e. if you compare the sbom.json in the artifacts you will see different dependencies)

erikbosch:
I get some weird errors in https://github.com/boschglobal/kuksa-databroker/actions/runs/8966502432/job/24622725710

i.e. when running https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/.github/workflows/create_draft_release.yml on this branch

```
Run which cross || cargo install cross
  which cross || cargo install cross
  cargo install cargo-license cargo-cyclonedx
  pip install "git+https://github.com/eclipse-kuksa/kuksa-common.git@6f3d7627760582d8ba83cc8a0f7449d00fffee84#subdirectory=sbom-tools"
  shell: /usr/bin/bash -e {0}
  env:
    CARGO_TERM_COLOR: always
/home/runner/.cargo/bin/cross
    Updating crates.io index
error: binary `cargo-license` already exists in destination
Add --force to overwrite
error: binary `cargo-cyclonedx` already exists in destination
Add --force to overwrite
     Summary Failed to install cargo-license, cargo-cyclonedx (see error(s) above).
error: some crates failed to install
Error: Process completed with exit code 101.
```

Could it possibly be that we use same cache key for databroker and databroker-cli? I will do some troubleshooting

`key: databroker-release-${{ matrix.platform.name }}-${{ hashFiles('**/Cargo.lock') }}
`

argerus:
> ```
> error: binary `cargo-license` already exists in destination
> Add --force to overwrite
> error: binary `cargo-cyclonedx` already exists in destination
> Add --force to overwrite
>      Summary Failed to install cargo-license, cargo-cyclonedx (see error(s) above).
> error: some crates failed to install
> Error: Process completed with exit code 101.
> ```
> 
> Could it possibly be that we use same cache key for databroker and databroker-cli? I will do some troubleshooting

No, but that is also not right.

As the message says the binaries are already installed (cached & restored) and trying to install them again will generate this error unless `--force` is added.

This doesn't happen for cargo cross, since it's only installed if the binary is missing with:
```
   which cross || cargo install cross
```

build-databroker.sh
argerus, line 1: Since we're (re)creating these scripts now anyway, can we move them into a dedicated `scripts/` folder?

SebastianSchildt:
> > ```
> > error: binary `cargo-license` already exists in destination
> > Add --force to overwrite
> > error: binary `cargo-cyclonedx` already exists in destination
> > Add --force to overwrite
> >      Summary Failed to install cargo-license, cargo-cyclonedx (see error(s) above).
> > error: some crates failed to install
> > Error: Process completed with exit code 101.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could it possibly be that we use same cache key for databroker and databroker-cli? I will do some troubleshooting
> 
> No, but that is also not right.
> 
> As the message says the binaries are already installed (cached & restored) and trying to install them again will generate this error unless `--force` is added.
> 
> This doesn't happen for cargo cross, since it's only installed if the binary is missing with:
> 
> ```
>    which cross || cargo install cross
> ```

Interesting. I only did observe this with cross, thus I used the "which pattern" I found in the workflows here.  I could use the same pattern for the other cargo tools, but question: Would we risk of running with "arbitrarily" old version for a long time, and once github throws cache suddenly get a new one?

Is there any way to always "upgrade" to a new stable version, if it exists? I think "--force" would always reinstall, that would work, but cost time...

build-databroker.sh
SebastianSchildt: This is out now. We are still deleting target/release in the build step (see comment in the script), so this might not save mich time currently

SebastianSchildt:
> Could it possibly be that we use same cache key for databroker and databroker-cli? I will do some troubleshooting
> 
> `key: databroker-release-${{ matrix.platform.name }}-${{ hashFiles('**/Cargo.lock') }} `

As @argerus  said, that is not the root cause, but has been fixed meanwhile

build-databroker.sh
SebastianSchildt: Ok I checked, the `rm -rf ./target/release` is still necessary, becasue

```
rm -rf target
rm -rf dist
./build-databroker.sh amd64
./build-databroker.sh riscv64
```

still yields

```
 Compiling bytes v1.6.0
error: failed to run custom build command for `libc v0.2.153`

Caused by:
  process didn't exit successfully: `/target/release/build/libc-71ad6c3e8bc9cca0/build-script-build` (exit status: 1)
  --- stderr
  /target/release/build/libc-71ad6c3e8bc9cca0/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /target/release/build/libc-71ad6c3e8bc9cca0/build-script-build)
  /target/release/build/libc-71ad6c3e8bc9cca0/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /target/release/build/libc-71ad6c3e8bc9cca0/build-script-build)
warning: build failed, waiting for other jobs to finish...
```

build-databroker.sh
SebastianSchildt: I will take a look. No promises. Don't want this PR to get a "bottomless pit", but maybe it is easy 🪦

erikbosch:
> > Could it possibly be that we use same cache key for databroker and databroker-cli? I will do some troubleshooting
> 
> No, but that is also not right.
> 
> As the message says the binaries are already installed (cached & restored) and trying to install them again will generate this error unless `--force` is added.
> 
> This doesn't happen for cargo cross, since it's only installed if the binary is missing with:
> 
> ```
>    which cross || cargo install cross
> ```

I got the impression that an error no longer shall be given upon reinstall based on the discussion in https://github.com/rust-lang/cargo/issues/6727.  When running cargo install locally I do not get any error when trying to install cargo-license even if it is already there

```
erik@debian6:~/kuksa-databroker$ cargo install cargo-license
    Updating crates.io index
erik@debian6:~/kuksa-databroker$ echo $?
0
erik@debian6:~/kuksa-databroker$ cargo install cargo-license
    Updating crates.io index
erik@debian6:~/kuksa-databroker$ echo $?
0

```

erikbosch:
I did some more tests on the `cargo install` issue. According to the links below `--force` shall not be needed. But it seems that we better should add two more files to the caching; ` ~/.cargo/.crates.toml` and `~/.cargo/.crates2.json`. With that https://github.com/boschglobal/kuksa-databroker/actions/workflows/create_draft_release.yml seems to work as expected.


```
  - uses: actions/cache@v4
      with:
        path: |
          ~/.cargo/bin/
          ~/.cargo/registry/index/
          ~/.cargo/registry/cache/
          ~/.cargo/git/db/
          ~/.cargo/.crates.toml
          ~/.cargo/.crates2.json
          target/
```

References:

- https://github.com/rust-lang/cargo/issues/11513
- https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci

SebastianSchildt:
@erikbosch  I just googled the same :D Will try. Still trying to move things....

.github/workflows/kuksa_databroker-cli_build.yml
SebastianSchildt: Yes, but that uses the old bom script, that does not take into account that SBOM is different, depending on the target architecture (i.e. it is only run once). The goal is to have an SBOM _exactly_ for a delivered binary.

No _theoretically_ it might be split (i.e. running this in parallel also for the 3 currently build architectures), but I think the benefits far outweigh the risks of mis-using this. Because you would need to "combine" the correct SBOMs with the correct builds. If you would use a different set of "feature" flags in cargo or potentially another rust version on another runner, this will not match. So in principle SBOMs should be generated "as close to the build" as possible. Currently by running them in the same script, they run exactly in the same environment as the build. And that is also true, when using this outside CI.

build-databroker.sh
SebastianSchildt: I was trying to make the script in a way, that it doesn't matter for where you call it (was thinking of manual use, which is what I did for testing locally)
(I could of course alternatively in the script cd to the main directory, but well, I didn't)

Any reason to change it again (it is a bit tedious)?

build-databroker.sh
argerus: But that is not the workflow when running in CI, so why should the CI build time be increased because of that..?

SebastianSchildt: target is in cache, so I assume the exact same can happen in CI, it just depends on the order of builds and what state of cache gets reinstantiated

build-databroker.sh
argerus: > I was trying to make the script in a way, that it doesn't matter for where you call it (was thinking of manual use, which is what I did for testing locally)

Yes, exactly.. Since it doesn't (shouldn't) matter what your working directory is when calling the script, why change the working directory etc? 

> Any reason to change it again (it is a bit tedious)?

I'm wondering why you went through the tedious task changing all the references from for example `dist/FILE` to `scripts/dist/FILE` which doesn't really make sense.. Revert that and change 1 line instead seems to make more sense?

erikbosch (dismissed):
Create Draft Release works now

https://github.com/boschglobal/kuksa-databroker/actions/workflows/create_draft_release.yml

SebastianSchildt:
thanks for testing @erikbosch , will make some minor adjustments tomorrow, and __maybe__ solve the target caching problem. I just think we really, really NEED correct(er), better standard compliant SBOMS, even though this now feels like refactoring whole build 🗡️  Well, as long as stuff gets better....

build-databroker.sh
argerus: Nope, since the cache key contains the platform, a build directory that was used for building a different platform / arch will never be pulled.

```yaml
key: databroker-release-${{ matrix.platform.name }}-${{ hashFiles('**/Cargo.lock') }}
```

erikbosch: But then `rm -rf ./target/release` does not have a negative impact on CI-builds, and is needed at least for local builds as shown by Sebastian, so no problem from a performance perspective to keep it, or?

argerus: > But then `rm -rf ./target/release` does not have a negative impact on CI-builds

What do you mean?

If you `rm -rf` the cache before building, the CI will obviously not be able to use what was in the cache.. that will impact the performance.

erikbosch: My mistake.

For my needs to be able to build locally it would be no problem to have an opt-in for clean, possibly something like below causing a clean of target/release before each target being built.

`./build-databroker.sh clean arm64 amd64`

erikbosch (approved):
LGTM

The release script works, tested at https://github.com/boschglobal/kuksa-databroker/releases

.github/workflows/kuksa_databroker-cli_build.yml
argerus: It would be good if generating the SBOM was at least a separate step.. Currently, I don't know whether it's that or something with the compilation that is causing the prolonged build time.

This PR:
Build time - PR


Baseline: 

Build time - Baseline

build-databroker.sh
SebastianSchildt: I think I found a way now to safely cache, that also works locally. You can tell cargo to use  different directories for target, I split now depending on architecture.

SebastianSchildt:
I would appreciate tif this to be merged now. This is quite large now, and the thing we NEED is sboms. 


Nothing against optimising it further (maybe make tickets), but we should move forward so I (or someone) don't need to port over all this stuff again in some weeks

erikbosch (approved):
Good enough, lets follow up in other PRs if needed

erikbosch:
Any objections to merging @argerus?

@SebastianSchildt - I approved but I believe you have to make sure that John approves or alternatively that you dismiss his review to be able to merge

argerus (approved):
> I would appreciate tif this to be merged now. This is quite large now, and the thing we NEED is sboms.
> 
> Nothing against optimising it further (maybe make tickets), but we should move forward so I (or someone) don't need to port over all this stuff again in some weeks

Sure. But, as I understand it, we already generate a Software Bill Of Materials. This is a different (more correct?) way of doing it. 

But it makes the compilation step take ~2x longer (?) which I think is not warranted. That can be restored in a different PR then I guess.

PR #26, opened 2024-05-06
Databroker performance improvements in Get service call

rafaeling:
This change means a **8 times faster** improvement for the Get method, which drops to **7.8 ms** for 5000 requests compared to the main branch implementation, which is **60 ms**.

See the wildcard exceptions that have been made obsolete by this change:
https://github.com/eclipse-kuksa/kuksa-databroker/blob/9ec27697af5d7594245e5c7cd5ebc2eb687abb80/doc/wildcard_matching.md

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -61,6 +61,7 @@ pub enum RegistrationError {
 pub struct Metadata {
     pub id: i32,
     pub path: String,
+    pub glob_path: String,
     pub data_type: DataType,
     pub entry_type: EntryType,
     pub change_type: ChangeType,
@@ -1155,6 +1156,7 @@ impl<'a, 'b> DatabaseWriteAccess<'a, 'b> {
             metadata: Metadata {
                 id: temp_id,
                 path: name.clone(),
+                glob_path: name.replace('.', "/"),
                 data_type,
                 change_type,
                 entry_type,
--- a/databroker/src/glob.rs
+++ b/databroker/src/glob.rs
@@ -11,12 +11,67 @@
 * SPDX-License-Identifier: Apache-2.0
 ********************************************************************************/
 
+use glob_match::glob_match;
 use lazy_static::lazy_static;
 use regex::Regex;
 
 #[derive(Debug)]
 pub enum Error {
     RegexError,
+}
+
+#[derive(Debug)]
+pub enum MatchError {
+    MatchError,
+}
+
+#[derive(Debug)]
+pub struct Matcher {
+    glob_pattern: String,
+}
+
+impl Matcher {
+    pub fn new(glob_pattern: &str) -> Result<Self, MatchError> {
+        // Check if the pattern is valid
+        if is_valid_pattern(glob_pattern) {
+            Ok(Matcher {
+                glob_pattern: Matcher::to_glob_string(glob_pattern),
+            })
+        } else {
+            Err(MatchError::MatchError)
+        }
+    }
+
+    pub fn is_match(&self, path: &str) -> bool {
+        glob_match(&self.glob_pattern, path)
+    }
+
+    pub fn as_string(&self) -> String {
+        self.glob_pattern.clone()
+    }
+
+    fn to_glob_string(glob: &str) -> String {
+        let mut glob_string = String::new();
+
+        if glob.eq("\"\"") || glob.is_empty() {
+            glob_string += "**";
+            return glob_string;
+        }
+
+        let mut glob_path_string = glob.replace('.', "/");
+
+        if !glob.contains('*') {
+            if glob.ends_with('.') {
+                glob_path_string += "/*";
+            } else if !glob.contains('.') {
+                glob_path_string += "/**";
+            }
+        }
+
+        glob_string.push_str(glob_path_string.as_str());
+
+        glob_string
+    }
 }
 
 pub fn to_regex_string(glob: &str) -> String {
@@ -69,6 +124,10 @@ pub fn to_regex_string(glob: &str) -> String {
 pub fn to_regex(glob: &str) -> Result<Regex, Error> {
     let re = to_regex_string(glob);
     Regex::new(&re).map_err(|_err| Error::RegexError)
+}
+
+pub fn glob_string_match(glob: &str, path: &str) -> bool {
+    glob_match(glob, path)
 }
 
 lazy_static! {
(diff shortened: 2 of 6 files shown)


SebastianSchildt:
This may be faster, but it breaks a lot in the existing API as well.

before
```

Test Client> getValue Vehicle.CurrentLocation 
[
    {
        "path": "Vehicle.CurrentLocation.GNSSReceiver.MountingPosition.X"
    },
    {
        "path": "Vehicle.CurrentLocation.Latitude"
    },
    {
        "path": "Vehicle.CurrentLocation.VerticalAccuracy"
    },
    {
        "path": "Vehicle.CurrentLocation.HorizontalAccuracy"
    },
    {
        "path": "Vehicle.CurrentLocation.Heading"
    },
    {
        "path": "Vehicle.CurrentLocation.Timestamp"
    },
    {
        "path": "Vehicle.CurrentLocation.GNSSReceiver.MountingPosition.Z"
    },
    {
        "path": "Vehicle.CurrentLocation.Longitude"
    },
    {
        "path": "Vehicle.CurrentLocation.GNSSReceiver.FixType"
    },
    {
        "path": "Vehicle.CurrentLocation.Altitude"
    },
    {
        "path": "Vehicle.CurrentLocation.GNSSReceiver.MountingPosition.Y"
    }
]

Test Client>       
```

After

```
Test Client> getValue Vehi
EXCEPTION of type 'Exception' occurred with message: 'Wrong databroker version, please use a newer version'
To enable full traceback, run the following command: 'set debug true'
Test Client> getValue Vehi
EXCEPTION of type 'Exception' occurred with message: 'Wrong databroker version, please use a newer version'
To enable full traceback, run the following command: 'set debug true'
Test Client> getValue Vehicle.CurrentLocation
{
    "error": {
        "code": 404,
        "reason": "not_found",
        "message": "Path not found"
    },
    "errors": [
        {
            "path": "Vehicle.CurrentLocation",
            "error": {
                "code": 404,
                "reason": "not_found",
                "message": "Path not found"
            }
        }
    ]
}

Test Client> 
```

(I guess breaking AutoComplete is another side effect of this)

So that would mean a major version bump and losing features

If RegExp (I guess the extending/matching with stars) is the performance bottle neck here, would it not be better, to only use RegExp if needed?

As in
 
```
if "*" in path do regexp path
else 
    doSameAsHere() and
          if leaf: return stuff
          if branch "regexp it"
         else "404"
´´´


That would just cost a bit of ifs, and I guess the heavy part of the regexps is, that you can't precompile (becasue they come form client). So implementing like above: A client that needs performance ONLY subscribes to Lead nodes and is fine,. Everybody else can knick themselves out

**Comment by rafaeling** [comment]
This was a hot fix to see if we could get some performance improvements.
We are still considering doing more research to keep the wildcard combined with good performance without these PR changes.
If the research results are not so promising, there is an additional idea of a new algorithm to partially support the wildcard only for the `.*` and `.**` cases while this PR O(1) algorithm.
I will change it to Draft as it should not be merged yet.

One more thing regarding auto-completion, yes, it is good to have it and it has been considered to implement an extra service method for Metadata that will support wildcards, which means that the user would not need wildcard support for Get as all signals could be stored locally by the client once in advance.

**Comment by SebastianSchildt** [comment]
Yes, in terms of an updated/new API I am sure there may be better ways to split/do this

However, for the current API, we should not merge something into main, that changes behavior significantly (I mean this doesn"t reinterpret some unclarities, it would actually remove features contradict what is documented)

"Engineering" wise,I am not sure, but have you looked into whether "globbing" libraries are faster? Like i.e.  https://github.com/devongovett/glob-match , I however lack the background to know if this is "fundamentally" faster then the regexp construction of state amchie and machting. And it also wil be a bandaid, because non-wildcard hashmap lookup shall be O(1) (with respect of number of signals, the hashing obviously would be O(l) with l being length of the path), and I doubt the most clever glob matching when backed by a "normal" map/list can ever be faster than O(n)..... But surprise me, if you can deliver O(1) we will probably take it  :D

so tldr: I still think now and also in a future  APIs ome kind of matching useful, but should be implemented in way that it is only "used when needed"


Off topic here, but since you mentioned it: I guess even when "discoverability" is in a metadata API a user might still want to subscribe to "a branch" (which can be seen as a "wildcard" sort of), otherwise something like the recorder from https://github.com/eclipse-kuksa/kuksa-csv-provider gets really hard, consider you want to generate a trace of everzthing running throug databroker (or a subbranch) but are not able to say "subscribe vehicle", but instead need to list/do 1000 individual paths

**Comment by rafaeling** [comment]
Trying out the library you posted, will test it and check the performance results.

**Comment by SebastianSchildt** [comment]
I think, that sutff like

`**.*.*.*.Position` is not supported anymore is fine, does not seem very real world, however is something preventing us to keep this

`Vehicle.Cabin.Sunroof` working as before? I have not checked the implementation, but might something like "if isBranch ad .**" make sense somewhere?

Rationale is: the first case I have never used, and find it hard to believe anyone did, the second one I definitely use often at least via CLI

File: databroker/src/broker.rs [hunk 1]
```diff
@@ -1155,6 +1156,7 @@ impl<'a, 'b> DatabaseWriteAccess<'a, 'b> {
             metadata: Metadata {
                 id: temp_id,
                 path: name.clone(),
+                glob_path: name.replace('.', "/"),
```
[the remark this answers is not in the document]
rafaeling: Actually we do care. All clients are configured to use `.` as the path request separator instead of `/`. 
If the `glob_path` string is not added during initialization, it would need to interate over all the signals, replace `.` with `/`, and then checking if each signal matches the request path.

By including the `glob_path` string during initialization, the process is streamlined: it simply translates the request path to the corresponding glob matching path and use it directly. This eliminates the need for repeated replacements and comparisons for each request.

rafaeling:
> I think, that sutff like
> 
> `**.*.*.*.Position` is not supported anymore is fine, does not seem very real world, however is something preventing us to keep this
> 
> `Vehicle.Cabin.Sunroof` working as before? I have not checked the implementation, but might something like "if isBranch ad .**" make sense somewhere?
> 
> Rationale is: the first case I have never used, and find it hard to believe anyone did, the second one I definitely use often at least via CLI

`Vehicle.Cabin.Sunroof` is not supported anymore, instead `Vehicle.Cabin.Sunroof.**` or `Vehicle.Cabin.Sunroof.*` should be use.
As far as I understand, there is no way in Databroker to determine if a path is a branch or not; we can only query whether a path is contained in the Databroker database or not. Previously, it was supported because Regex does not have as many limitations as the glob matching library approach.

databroker/src/glob.rs
```diff
@@ -19,6 +19,29 @@ pub enum Error {
     RegexError,
 }
 
+pub fn to_glob_string(glob: &str) -> String {
```
argerus, line 22: I think it would be better to have not return a "glob string" from `crate::glob`. Instead, the implementation, i.e. the usage of `glob_match` or such implementation detail should be handled inside this file and only a `glob::Matcher` or something equivalent should be returned for use by external users.

PS. The same thing applies to `to_regex_string`

databroker/src/glob.rs
```diff
@@ -19,6 +19,29 @@ pub enum Error {
...
+    let mut re = String::new();
+
+    if glob.is_empty() {
+        re += "**";
+        return re;
+    }
+
+    let mut regex_string = glob.replace('.', "/");
```
argerus, line 30:
```suggestion
    let mut glob_string = glob.replace('.', "/");
```

databroker/src/glob.rs
```diff
@@ -19,6 +19,29 @@ pub enum Error {
     RegexError,
 }
 
+pub fn to_glob_string(glob: &str) -> String {
+    let mut re = String::new();
```
argerus, line 23: not a `re`

databroker/src/glob.rs
```diff
@@ -136,919 +159,919 @@ mod tests {
...
-        "Vehicle.VehicleIdentification.VehicleSeatingCapacity",
-        "Vehicle.VehicleIdentification.VehicleSpecialUsage",
-        "Vehicle.VehicleIdentification.WMI",
-        "Vehicle.VehicleIdentification.Year",
-        "Vehicle.VersionVSS.Label",
-        "Vehicle.VersionVSS.Major",
-        "Vehicle.VersionVSS.Minor",
-        "Vehicle.VersionVSS.Patch",
-        "Vehicle.Width",
+        "Vehicle/_kuksa/databroker/GitVersion",
```
argerus, line 162: We are still using the regex matching (for tokens), so the  tests for that should not be removed. Or in other words, the tests should be removed when the code they are testing is removed.

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -375,50 +366,49 @@ impl proto::val_server::Val for broker::DataBroker {
...
-        let mut valid_requests: HashMap<String, (regex::Regex, HashSet<broker::Field>)> =
-            HashMap::new();
+        let mut valid_requests: HashMap<String, (String, HashSet<broker::Field>)> = HashMap::new();
 
         for entry in &request.entries {
-            if entry.path.contains('*') && !glob::is_valid_pattern(&entry.path) {
+            if (entry.path.contains('*') && !glob::is_valid_pattern(&entry.path))
+                || entry.path.len() > MAX_REQUEST_PATH_LENGTH
+            {
                 tonic::Status::new(tonic::Code::InvalidArgument, "Invalid Pattern Argument");
```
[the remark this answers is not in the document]
rafaeling: Done

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -72,7 +75,9 @@ impl proto::val_server::Val for broker::DataBroker {
 
             // Fill valid_requests structure.
             for request in requested {
-                if request.path.contains('*') && !glob::is_valid_pattern(&request.path) {
+                if (request.path.contains('*') && !glob::is_valid_pattern(&request.path))
+                    || request.path.len() > MAX_REQUEST_PATH_LENGTH
```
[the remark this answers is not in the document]
rafaeling: Done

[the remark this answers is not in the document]
argerus: Doesn't it make more sense to check for a valid pattern when creating the glob matcher, i.e. the proposed `glob::Matcher`, or if we're not doing that, when creating the "glob string".

argerus (changes requested):

databroker/src/broker.rs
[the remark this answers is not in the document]
rafaeling: I think we need both in case we don't want to make so many changes to the code.
So in case Databroker stores paths **only** with `/` to be able to support the wildcard case, it would mean that all grpc / viss call layer must be alligned to convert the path requested to use `/`.
 
The purpose of adding this `glob_path` string is to use glob_match library **only** for get and subscribe methods (wildcard use cases), meaning we dont have to change and adapt all grpc /viss calls just to solve this specific wildcard use case.

databroker/src/grpc/kuksa_val_v1/val.rs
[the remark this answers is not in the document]
rafaeling: `is_valid_pattern` signature I think is not correct, would be better if `is_valid_path`.
Here it should check if path requested by the user is valid, not if the glob string is valid.

databroker/src/glob.rs
rafaeling: Will add them back

SebastianSchildt:
> > I think, that sutff like
> > `**.*.*.*.Position` is not supported anymore is fine, does not seem very real world, however is something preventing us to keep this
> > `Vehicle.Cabin.Sunroof` working as before? I have not checked the implementation, but might something like "if isBranch ad .**" make sense somewhere?
> > Rationale is: the first case I have never used, and find it hard to believe anyone did, the second one I definitely use often at least via CLI
> 
> `Vehicle.Cabin.Sunroof` is not supported anymore, instead `Vehicle.Cabin.Sunroof.**` or `Vehicle.Cabin.Sunroof.*` should be use. As far as I understand, there is no way in Databroker to determine if a path is a branch or not; Databroker can only internally query whether a path is contained in the database or not. Previously, it was supported because Regex does not have as many limitations as the glob matching library approach.

Ah, I remember, because "branches" are not really a thing in our internal data structure. In terms of keeping API compatibility, would something be feasible like

```
inputstr=`Vehicle.Cabin.Sunroof`

res = glob_the_shit_out_of_this(inputStr)

if res=no match
   res = glob_the_shit_out_of_this(inputStr+".**")
```

It should not really cost anything , when querying direct leaves.

I understand, for any kind of new API we would always use  the "best" approch, but this is slightly modifying existing API behavior, so I would prefer to keep behavior changes to a minimum

databroker/src/glob.rs
rafaeling: DOne

databroker/src/glob.rs
rafaeling: Done, all is handled now in the glob file.

argerus: It's still returning a "glob string" which is just a `String`.
And then using `glob_string_match(glob: &str, path: &str)` on that "glob string".

My point was that it would be better to have an API like this:
```rust
// in glob.rs

impl Matcher {
  pub fn new(glob: &str) -> Result<Matcher, Error> {

  }

  pub fn is_match(path: &str) {

  }
}
```

so that we don't pass things called "glob strings" around that are no different than `String`s.

That would make it easier to use (and harder to accidentally get the strings wrong) since it would be type safe.

rafaeling:
> > > I think, that sutff like
> > > `**.*.*.*.Position` is not supported anymore is fine, does not seem very real world, however is something preventing us to keep this
> > > `Vehicle.Cabin.Sunroof` working as before? I have not checked the implementation, but might something like "if isBranch ad .**" make sense somewhere?
> > > Rationale is: the first case I have never used, and find it hard to believe anyone did, the second one I definitely use often at least via CLI
> > 
> > 
> > `Vehicle.Cabin.Sunroof` is not supported anymore, instead `Vehicle.Cabin.Sunroof.**` or `Vehicle.Cabin.Sunroof.*` should be use. As far as I understand, there is no way in Databroker to determine if a path is a branch or not; Databroker can only internally query whether a path is contained in the database or not. Previously, it was supported because Regex does not have as many limitations as the glob matching library approach.
> 
> Ah, I remember, because "branches" are not really a thing in our internal data structure. In terms of keeping API compatibility, would something be feasible like
> 
> ```
> inputstr=`Vehicle.Cabin.Sunroof`
> 
> res = glob_the_shit_out_of_this(inputStr)
> 
> if res=no match
>    res = glob_the_shit_out_of_this(inputStr+".**")
> ```
> 
> It should not really cost anything , when querying direct leaves.
> 
> I understand, for any kind of new API we would always use the "best" approch, but this is slightly modifying existing API behavior, so I would prefer to keep behavior changes to a minimum


I just implemented what you suggested.
But for this use case `"**.Sunroof"` will add ".**" and will return four signals.

Previously this use case should have returned nothing (see https://github.com/eclipse-kuksa/kuksa-databroker/blob/bd14dcd850bbe82a65c563e13d37d1eb45d42059/doc/wildcard_matching.md)

rafaeling:
> > > I think, that sutff like
> > > `**.*.*.*.Position` is not supported anymore is fine, does not seem very real world, however is something preventing us to keep this
> > > `Vehicle.Cabin.Sunroof` working as before? I have not checked the implementation, but might something like "if isBranch ad .**" make sense somewhere?
> > > Rationale is: the first case I have never used, and find it hard to believe anyone did, the second one I definitely use often at least via CLI
> > 
> > 
> > `Vehicle.Cabin.Sunroof` is not supported anymore, instead `Vehicle.Cabin.Sunroof.**` or `Vehicle.Cabin.Sunroof.*` should be use. As far as I understand, there is no way in Databroker to determine if a path is a branch or not; Databroker can only internally query whether a path is contained in the database or not. Previously, it was supported because Regex does not have as many limitations as the glob matching library approach.
> 
> Ah, I remember, because "branches" are not really a thing in our internal data structure. In terms of keeping API compatibility, would something be feasible like
> 
> ```
> inputstr=`Vehicle.Cabin.Sunroof`
> 
> res = glob_the_shit_out_of_this(inputStr)
> 
> if res=no match
>    res = glob_the_shit_out_of_this(inputStr+".**")
> ```
> 
> It should not really cost anything , when querying direct leaves.
> 
> I understand, for any kind of new API we would always use the "best" approch, but this is slightly modifying existing API behavior, so I would prefer to keep behavior changes to a minimum

Implemented your suggestions. Now it is also compatible for branches use cases like `Vehicle.Cabin.Sunroof`
Please test it, meanwhile I will try fix the format errors as soon as possible.

SebastianSchildt:
@rafaeling  sounds good

It would be good, if the conflicts are resolved, because that seems to prevent CI runs, so currently  can not lazily test by just pulling the docker image :D

SebastianSchildt:
Built manually and checked behavior. Looking good to me!

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -72,7 +75,17 @@ impl proto::val_server::Val for broker::DataBroker {
...
             // Fill valid_requests structure.
             for request in requested {
-                if request.path.contains('*') && !glob::is_valid_pattern(&request.path) {
+                if request.path.len() > MAX_REQUEST_PATH_LENGTH {
+                    errors.push(proto::DataEntryError {
+                        path: request.path,
+                        error: Some(proto::Error {
+                            code: 400,
+                            reason: "bad_request".to_owned(),
+                            message: "Invalid Request Path Length".to_owned(),
```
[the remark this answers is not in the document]
rafaeling: Done

databroker/src/grpc/kuksa_val_v1/val.rs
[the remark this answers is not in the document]
rafaeling: Since we use request paths with point separators and asterisks in wildcards for all the kuksa clients would not be enough for now to check just if the request path is a valid request pattern?.
Otherwise I have to implement a new glob string regex (last time took me long time). 
`is_valid_pattern` as requested path "filter" for now is probably enough?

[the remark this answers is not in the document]
argerus: > Since we use request paths with point separators and asterisks in wildcards for all the kuksa clients would not be enough for now to check just if the request path is a valid request pattern?

I don't think I understand what you mean.. And I guess I was a bit unclear as well.

What I'm saying is:

1. Why is the path only validated if the path contains a `*`. Shouldn't the path always be validated?
2. Why is `glob::is_valid_pattern` not part of `glob::Matcher::new()`. It should not be possible to create a glob matcher with an invalid path, so it should be part of the constructor. Right now, the user has to remember to do the validation. It's better to have the Rust type system guarantee this.

>  Otherwise I have to implement a new glob string regex (last time took me long time).

Why? The validation function is already there.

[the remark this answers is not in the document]
rafaeling: Got it now, introduced the Matcher struct that wraps that functionality, you can take a look at it

databroker/src/glob.rs
```diff
@@ -19,6 +20,69 @@ pub enum Error {
...
+        !self.glob_pattern.starts_with("**") && !self.glob_pattern.ends_with("/**")
+    }
+
+    pub fn glob_match_branch(&self, path: &str) -> bool {
+        let glob_match_branch = format!("{}/**", self.glob_pattern);
+        glob_match(&glob_match_branch, path)
+    }
+
+    pub fn get_string(&self) -> String {
+        self.glob_pattern.clone()
```
argerus, line 59:
```suggestion
    pub fn as_string(&self) -> String {
        self.glob_pattern.clone()
```

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -143,17 +148,56 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                        .await;
+
+                    // Not found any matches meaning it could be a branch path request
+                    if matcher.is_branch_matcher() && !(*is_match) {
+                        //*glob += "/**";
+                        broker
+                            .for_each_entry(|entry| {
+                                let mut result_fields: HashSet<proto::Field> = HashSet::new();
+                                let glob_path = &entry.metadata().glob_path;
+                                if matcher.glob_match(glob_path) {
```
argerus, line 168: Isn't this:
```rust
if matcher.glob_match(glob_path) {
```

doing the exact same thing... again?

And why do this only if the first iteration didn't match. We are doing a glob match anyway. Can we not match both branches and not branches with a single pattern at the same time?

argerus (changes requested):

databroker/src/glob.rs
```diff
@@ -19,6 +20,73 @@ pub enum Error {
...
+    }
+
+    pub fn transform_pattern_path_to_branch(&mut self) {
+        self.glob_pattern += "/**";
+    }
+
+    pub fn is_branch_match(&self, path: &str) -> bool {
+        glob_match(&self.glob_pattern, path)
+    }
+
```
argerus, line 61: Does `is_branch_match(...)` make any sense any more? It looks like it's just an historical artifact that should be removed.

databroker/src/glob.rs
```diff
@@ -19,6 +20,73 @@ pub enum Error {
...
+    // Only support branches like Vehicle.Cabin.Sunroof but not like **.Sunroof
+    pub fn is_branch_matcher(&self) -> bool {
+        !self.glob_pattern.starts_with("**") && !self.glob_pattern.ends_with("/**")
+    }
+
+    pub fn transform_pattern_path_to_branch(&mut self) {
+        self.glob_pattern += "/**";
+    }
```
argerus, line 56: Does `transform_pattern_path_to_branch` really belong in glob.rs. It's seems to be a very use case specific function that is better handled at the call site. If we want to match using a different glob, just create a new one:
```rust
glob::Matcher::new(old_pattern + "/**")
```

databroker/src/glob.rs
```diff
@@ -19,6 +20,69 @@ pub enum Error {
...
+        }
+    }
+
+    pub fn glob_match(&self, path: &str) -> bool {
+        glob_match(&self.glob_pattern, path)
+    }
+
+    pub fn is_branch_matcher(&self) -> bool {
+        !self.glob_pattern.starts_with("**") && !self.glob_pattern.ends_with("/**")
+    }
```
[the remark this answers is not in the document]
argerus: I still don't really understand the logic of this function.
Maybe it makes sense in the context from where it's called. But in that case, it should probably not be part of `glob::Matcher`, which is rather generic.

argerus (changes requested):

argerus (approved):

PR #29, opened 2024-05-09
Fix caching on databroker build

SebastianSchildt:
There was a typo that prevents full caching of build artifacts for databroker build (CLI was ok already). Fixed now. 

Example build with no cache (so probably also the build here, as changing cache contents invalidates cache)

Screenshot 2024-05-09 at 13 43 53

Subsequent build satisfying cache requirements

Screenshot 2024-05-09 at 13 47 12

(see here https://github.com/boschglobal/kuksa-databroker/actions?query=event%3Apull_request+branch%3Atest%2Fcaching for various test builds)

I also checked where time is mainly used now

 - 20+   sec is always used pulling the cross images (crossb uilds using docker). the cross _cargo_ package is cached and not reinstalled but the docker images are not. **I do not intend to look into this/fix this in this PR**. Just a hint, if somebody feels bored. I am however not sure if there is even a feasible solution as those images are around 2 GB each and I think there are strict limitations on how much cache can be used before things get evicted)
 - Another large chunk is basically the "linking" so I think, this has to be accepted.

### DIFF
--- a/.github/workflows/kuksa_databroker_build.yml
+++ b/.github/workflows/kuksa_databroker_build.yml
@@ -80,7 +80,7 @@ jobs:
             ~/.cargo/.crates.toml
             ~/.cargo/.crates2.json
             ~/.cache/pip/
-            target-*/
+            target/
           key: databroker-coverage-${{ hashFiles('**/Cargo.lock') }}
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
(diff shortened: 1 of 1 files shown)


argerus:
This seems to fix caching for the build step.

The overall build is still slower than it was since caching of the unit test step was also broken in the original PR.

Screenshot from 2024-05-10 14-41-25

I suppose always having the target dir under `target/` would make it easier to get this right, i.e. using `--target-dir "target/cross-$target"` in order to work around the issue with building for different systems under `cross`.

argerus (changes requested):
This seems to fix caching for the build step.

The overall build is still slower than it was since caching of the unit test step was also broken in the original PR.

Screenshot from 2024-05-10 14-41-25

I suppose always having the target dir under `target/` would make it easier to get this right, i.e. using `--target-dir "target/cross-$target"` in order to work around the issue with building for different systems under `cross`.

SebastianSchildt:
Fixed. I choose to fix the cache definition in workflow instead of changing dirs in script.
(I was thinking if somebody during dev is using plain "cargo build/run" and also in between places with the build scripts, the structure inside the default target folder might seem confusing. Also... easier/faster to test for me just now)

argerus (approved):

PR #30, opened 2024-05-10
Upgrade check-dash action to v4: Faster + auto create tickets

SebastianSchildt:
This allows dash to create tickets in Eclipses IP tracker if necessary

It is also faster then the old version, as installation time for JVM is greatly reduced

Before

image

After

image


erikbosch (approved):
LGTM

PR #31, opened 2024-05-13
Upgrading codecov dependency

erikbosch:
Seems to be working. Well - upload does not work due to a limit for tokenless update, likely same problem as we often see for v3. Will do a retry tomorrow morning when it maybe is lower load as the Americans are sleeping.

_It is MAYBE so that we could add a token to the fork, but we cannot expect that all forks have a suitable token, and it seems to imply that we enable codecov for that particular organization_

`
error - 2024-05-13 14:46:53,425 -- Report creating failed: {"detail":"Tokenless has reached GitHub rate limit. Please upload using a token: https://docs.codecov.com/docs/adding-the-codecov-token. Expected available in 924 seconds."}`

### DIFF
--- a/.github/workflows/kuksa_databroker_build.yml
+++ b/.github/workflows/kuksa_databroker_build.yml
@@ -87,11 +87,11 @@ jobs:
       - name: Generate code coverage
         run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
       - name: Upload coverage to Codecov
-        # To use v4 a token must be specifed by "token: ${{ secrets.CODECOV_TOKEN }}""
         # Uploaded result available at https://app.codecov.io/gh/eclipse-kuksa/kuksa-databroker
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           files: lcov.info
+          token: ${{ secrets.CODECOV_TOKEN }}
 
   build:
     name: Build


SebastianSchildt:
Ok, just saw the error, but let me summarize what we see here

 - For PRs like this the secret in our orga is not available, thus the Action tries "tokenless"
 - It may likely hit the rate limit, and print a message in logs, but the CI is still green (which I think it ok)
 - If we would merge something in main, of course CI runs in context ouf our orga, and thus Uplaod should work

If we want more regular suceeding upload (and leave a bit of bandwidth for outside contributors), we could add a similarly named token to the orga we mostly do PRs from, and it would to tokenized upload

Anyway, as I have seen the repo idle for a bit, I will try manually retriggering the run, to see what happens when it can upload


Maybe in the long run, there might be a way, where the PR only generated Codevov input file and saves it as artefact, and we have another workflow running on something like

```
on:
  workflow_run:
    workflows: ["CodeCov prep"]  # List of workflows to wait for
    types:
      - completed
```

But anyway, no idea that is working, more. logn term idea (this would need somehow to identify the run in question, find the artifct AND give Codecove all neccessary meta data, sommaybe not even possible)

SebastianSchildt (approved):
Seems to work as expected

Just maybe we want to configure/fine-tune  on what conditions this reports failure

erikbosch:
@SebastianSchildt - what we possibly should follow up is what is reported as "base". In the image below it uses a quite old version (17 commits!) , possibly because it was that long time ago we last had a successful upload for main.

image

PR #32, opened 2024-05-14
Adding subscription_id and stream_response_time_ns to subscribe response metadata [ --feature stats ]

LikhithST:
Adding metadata related to "time to process requests" for set and subscribe calls, adding under the feature stats


databroker/src/broker.rs
```diff
@@ -153,17 +153,49 @@ pub struct QuerySubscription {
     permissions: Permissions,
 }
 
+#[cfg(not(feature="stats"))]
```
erikbosch, line 156: If intended for main branch, should we possibly somewhere document what the feature "stats" is intended for and how it works.

argerus (changes requested):
So first of all, this change is way too intrusive for what I'm guessing is the intended purpose, so a short description of what the PR intends to accomplish would be good.

Secondly, if the intent is to collect metrics from a running databroker instance, have you looked into using the [`tracing`](https://docs.rs/tracing/latest/tracing/) library (which databroker already uses for logging) in order to collect these things and wire it up to open telemetry or something in order to subscribe to it?

data/vss-core/vss_release_4.0.json
```diff
@@ -1 +1 @@
...
...pdates-j3016-automated-driving-graphic taxonomy. For SAE levels 3 and 4 the system is required to alert the driver before it will disengage. Level 4 systems are required to reach a safe state even if a driver does not take over. Only level 5 systems are required to not rely on a driver at all. While level 2 systems require the driver to be monitoring the system at all times, many level 2 systems, often termed \"level 2.5\" systems, do warn the driver shortly before reaching their operational limits, therefore we also support the DISENGAGING state for SAE_2.", "datatype": "string", "description": "Indicates the currently active level of autonomy according to SAE J3016 taxonomy.", "type": "sensor", "uuid": "b101c6928fc55948b1cc485e568ecd8d" }, "CruiseControl": { "children": { "IsActive": { "datatype": "boolean", "description": "Indicates if cruise control system is active (i.e. actively controls speed). True = Active. False = Inactive.", "type": "actuator", "uuid": "78ab5ce923dc5aa1a6622bcb948e1561" }, "IsEnabled": { "datatype": "boolean", "description": "Indicates if cruise control system is enabled (e.g. ready to receive configurations and settings) True = Enabled. False = Disabled
```
erikbosch, line 1: What changes have you done here?

dockerfile
```diff
@@ -0,0 +1,16 @@
+# Use the latest version of the Rust base image
```
erikbosch, line 1: I find a bit confusing to have one dockerfile here and others in https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/scripts/Dockerfile

What is the difference between them. Should better be described somewhere.

Latency_Extractor_ReadMe.md
```diff
@@ -0,0 +1,49 @@
...
+
+This command will stop and remove the containers defined in the Docker Compose configuration.
+
+## Additional Information
+
+[Kuksa Data Broker Github Repo](https://github.com/eclipse-kuksa/kuksa-databroker).
+
+---
+
+By following these steps, you will be able to build and run the Kuksa Data Broker with the stats feature enabled. If you encounter any issues or have further questions, please refer to the official documentation or seek support from the community.
```
erikbosch, line 49: If we are to add a stats feature we should better have a better description on what it does. Like what stats does it collect and how can you review them. I would suggest adding some screenshots - it seems that we start a grafan instance, what would you be able to see there and so on.

Latency_Extractor_ReadMe.md
```diff
@@ -0,0 +1,49 @@
+
+# Kuksa Data Broker
```
erikbosch, line 2: If we are to merge this PR I would like a link from the main README to this file

erikbosch:
There are conflicts that needs to be fixed

lukasmittag:
Hey @LikhithST, for now I'm closing this. As discussed we will work on a fork on this. We then can decide to open a PR with some things that we also want here. @erikbosch thanks for taking a look, without context this PR can be confusing. @LikhithST thanks for the work you have done so far.

PR #34, opened 2024-05-24
Add StreamedUpdate API

wba2hi:
Add support for Streamed Update API. 

I know this is probably not yet in a perfect state, but I still want to trigger the first review. According to the ticket adapting the databroker-cli and databroker-providers is out-of-scope since this is only an intermediate solution and will be shortly overhauled by kuksa val v2 api anyway. I shortly tried to get it to work, but I failed with some weird error messages and incompatible steps. If this is an easy step I am still willing to do it, but I probably need a helping hand to support me there :) 

However, I verified the functionality using 
a) ghz (https://github.com/mikehaller/kuksa-databroker-ghz/tree/main) using the following testcase: 

It sends out 5 requests which 100_000 streamed data points per request
```
{
    "name": "Test Case 5: Kuksa API, StreamedUpdate",
    "proto": "kuksa/val/v1/val.proto",
    "insecure": true,
    "skipFirst": 1,
    "call": "kuksa.val.v1.VAL.StreamedUpdate",
    "total": 6,
    "disable-template-functions": true,
    "disable-template-data": true,
    "stream-call-count": 100000,
    "concurrency": 1,
    "data": {
        "updates": [    
            {
                "entry": {
                    "path": "Vehicle.Speed",
                    "value": {
                        "float": 100.0
                    }
                },
                "fields": [2]
            }
        ]
    }
}
```
using command `ghz --config=testcase105_kuksa_testcase105_kuksa_streamedupdate.json --import-paths=/Users/wba2hi/git/kuksa-databroker/proto localhost:55556 > testcase105_kuksa_streamedupdate.json.output`

and b)
a prototypical implementation in android: https://github.com/boschglobal/kuksa-android-sdk/tree/user/wba2hi/VAL-1320

As an orientation help I used the stream_datapoints api from collector.rs

### DIFF
--- a/databroker/src/grpc/kuksa_val_v1/val.rs
+++ b/databroker/src/grpc/kuksa_val_v1/val.rs
@@ -840,6 +840,7 @@ impl broker::EntryUpdate {
 
 #[cfg(test)]
 mod tests {
+    use databroker_proto::kuksa::val::v1::val_server::Val;
     use super::*;
     use crate::{broker::DataBroker, permissions};
 
@@ -895,6 +896,100 @@ mod tests {
                 assert_eq!(error.code, 400, "unexpected error code");
             }
             Err(_status) => panic!("failed to execute set request"),
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn test_streamed_update_with_valid_datapoint() {
+        let broker = DataBroker::default();
+        let authorized_access = broker.authorized_access(&permissions::ALLOW_ALL);
+
+        authorized_access
+            .add_entry(
+                "Vehicle.Speed".to_owned(),
+                broker::DataType::Float,
+                broker::ChangeType::OnChange,
+                broker::EntryType::Sensor,
+                "Test datapoint 1".to_owned(),
+                None,
+                Some("km/h".to_owned()),
+            )
+            .await
+            .expect("Register datapoint should succeed");
+
+        let streamed_update_request = proto::StreamedUpdateRequest {
+            updates: vec![
+                proto::EntryUpdate {
+                    fields: vec![proto::Field::Value as i32],
+                    entry: Some(proto::DataEntry {
+                        path: "Vehicle.Speed".to_owned(),
+                        value: Some(proto::Datapoint {
+                            timestamp: Some(std::time::SystemTime::now().into()),
+                            value: Some(proto::datapoint::Value::Float(120.0)),
+                        }),
+                        metadata: None,
+                        actuator_target: None,
+                    }),
+                }],
+        };
+
+        let mut streaming_request = tonic_mock::streaming_request(vec![streamed_update_request]);
+        streaming_request.extensions_mut().insert(permissions::ALLOW_ALL.clone());
+        match broker.streamed_update(streaming_request).await {
+            Ok(response) => {
+                tokio::spawn(async move {
+                    let stream = response.into_inner();
+                    let mut receiver = stream.into_inner();
+                    let option = receiver.recv();
+                    assert!(option.await.is_none()) // no errors should occur and no ack is delivered
+                });
+            }
+            Err(_) => {
+                panic!("Should not happen")
+            }
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn test_streamed_update_with_invalid_datapoint() {
+        let broker = DataBroker::default();
+
+        let streamed_update_request = proto::StreamedUpdateRequest {
+            updates: vec![
+                proto::EntryUpdate {
+                    fields: vec![proto::Field::Value as i32],
+                    entry: Some(proto::DataEntry {
+                        path: "Vehicle.Invalid.Speed".to_owned(),
+                        value: Some(proto::Datapoint {
+                            timestamp: Some(std::time::SystemTime::now().into()),
+                            value: Some(proto::datapoint::Value::Float(120.0)),
+                        }),
+                        metadata: None,
+                        actuator_target: None,
+                    }),
+                }],
+        };
+
+        let mut streaming_request = tonic_mock::streaming_request(vec![streamed_update_request]);
+        streaming_request.extensions_mut().insert(permissions::ALLOW_ALL.clone());
+        match broker.streamed_update(streaming_request).await {
+            Ok(response) => {
+                tokio::spawn(async move {
+                    let stream = response.into_inner();
+                    let mut receiver = stream.into_inner();
+                    let option = receiver.recv().await;
+                    assert!(option.is_some());
+                    let result = option.unwrap();
+                    let error_opt = result.unwrap().error;
+                    let error = error_opt.unwrap();
+                    assert_eq!(error.code, 404);
+                    assert_eq!(error.reason, "not_found");
+                    assert_eq!(error.message, "Vehicle.Invalid.Speed not found")
+                });
+            }
+            Err(_) => {
+                panic!("Should not happen")
+            }
         }
     }
 }
--- a/databroker/src/grpc/kuksa_val_v1/val.rs
+++ b/databroker/src/grpc/kuksa_val_v1/val.rs
@@ -20,16 +20,16 @@ use tokio::sync::mpsc;
 
 use databroker_proto::kuksa::val::v1 as proto;
 use databroker_proto::kuksa::val::v1::{DataEntryError, EntryUpdate};
+use tokio_stream::wrappers::ReceiverStream;
 use tokio_stream::Stream;
 use tokio_stream::StreamExt;
-use tokio_stream::wrappers::ReceiverStream;
 use tonic::{Response, Status, Streaming};
 use tracing::debug;
 
 use crate::broker;
-use crate::broker::{AuthorizedAccess, EntryReadAccess};
 use crate::broker::ReadError;
 use crate::broker::SubscriptionError;
+use crate::broker::{AuthorizedAccess, EntryReadAccess};
 use crate::glob;
 use crate::permissions::Permissions;
 
@@ -225,7 +225,6 @@ impl proto::val_server::Val for broker::DataBroker {
 
         let broker = self.authorized_access(&permissions);
 
-
         let entry_updates = request.into_inner().updates;
 
         // Collect errors encountered
@@ -235,14 +234,10 @@ impl proto::val_server::Val for broker::DataBroker {
         for request in entry_updates {
             match &request.entry {
                 Some(entry) => match broker.get_id_by_path(&entry.path).await {
-                    Some(id) => {
-                        match validate_entry_update(&broker, &request, id).await {
-                            Ok(pair) => {
-                                updates.push((pair.first, pair.second))
-                            }
-                            Err(e) => { return Err(e) }
-                        }
-                    }
+                    Some(id) => match validate_entry_update(&broker, &request, id).await {
+                        Ok(pair) => updates.push((pair.first, pair.second)),
+                        Err(e) => return Err(e),
+                    },
                     None => {
                         let message = format!("{} not found", entry.path);
                         errors.push(proto::DataEntryError {
@@ -283,7 +278,8 @@ impl proto::val_server::Val for broker::DataBroker {
         }))
     }
 
-    type StreamedUpdateStream = ReceiverStream<Result<proto::StreamedUpdateResponse, tonic::Status>>;
+    type StreamedUpdateStream =
+        ReceiverStream<Result<proto::StreamedUpdateResponse, tonic::Status>>;
 
     async fn streamed_update(
         &self,
@@ -419,10 +415,10 @@ impl proto::val_server::Val for broker::DataBroker {
 
     type SubscribeStream = Pin<
         Box<
-            dyn Stream<Item=Result<proto::SubscribeResponse, tonic::Status>>
-            + Send
-            + Sync
-            + 'static,
+            dyn Stream<Item = Result<proto::SubscribeResponse, tonic::Status>>
+                + Send
+                + Sync
+                + 'static,
         >,
     >;
 
@@ -548,13 +544,16 @@ impl proto::val_server::Val for broker::DataBroker {
     }
 }
 
-async fn validate_entry_update(broker: &AuthorizedAccess<'_, '_>, request: &EntryUpdate, id: i32) -> Result<Pair<i32, broker::EntryUpdate>, Status> {
+async fn validate_entry_update(
+    broker: &AuthorizedAccess<'_, '_>,
+    request: &EntryUpdate,
+    id: i32,
+) -> Result<Pair<i32, broker::EntryUpdate>, Status> {
     let entry = &request.entry.clone().unwrap();
 
-    let fields =
-        HashSet::<proto::Field>::from_iter(request.fields.iter().filter_map(
-            |id| proto::Field::from_i32(*id), // Ignore unknown fields for now
-        ));
+    let fields = HashSet::<proto::Field>::from_iter(request.fields.iter().filter_map(
+        |id| proto::Field::from_i32(*id), // Ignore unknown fields for now
+    ));
 
     if entry.actuator_target.is_some() {
         if let Some(metadata) = broker.get_metadata(id).await {
@@ -568,16 +567,11 @@ async fn validate_entry_update(broker: &AuthorizedAccess<'_, '_>, request: &Entr
 
     let entry = match &request.entry {
         Some(entry) => entry,
-        None => {
-            return Err(tonic::Status::invalid_argument(
-                "Empty entry".to_string(),
-            ))
-        }
+        None => return Err(tonic::Status::invalid_argument("Empty entry".to_string())),
     };
 
     debug!("Setting fields: {:?}", fields);
-    let update =
-        broker::EntryUpdate::from_proto_entry_and_fields(entry, fields);
+    let update = broker::EntryUpdate::from_proto_entry_and_fields(entry, fields);
 
     let pair = Pair {
         first: id,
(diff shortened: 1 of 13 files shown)


databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -475,6 +548,100 @@ impl proto::val_server::Val for broker::DataBroker {
     }
 }
 
+async fn validate_entry_update(broker: &AuthorizedAccess<'_, '_>, request: &EntryUpdate, id: i32) -> Result<Pair<i32, broker::EntryUpdate>, Status> {
```
wba2hi, line 551: Since the streamed update api uses a lot of code from the set method I tried to separate some common things into separate methods (I did not want to redundantly copy 120 lines of code inside the loop mechanism).

I'm still not satisfied with the complexity from streamed_update method right now, but I still have some struggles with Rust and it's borrowing concept, making it hard for me to improve it even more. It also feels like a lot of complexity comes from all the matches and results?

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -344,12 +283,146 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                                        let path = metadata.path.clone();
+                                                        let data_entry_error = convert_to_data_entry_error(&path, &error);
+                                                        errors.push(data_entry_error);
+                                                    }
+                                                }
+
+                                                if let Err(err) = sender.send(
+                                                    Ok(proto::StreamedUpdateResponse {
+                                                        errors,
+                                                        error: None, // TODO take first element of errors
```
wba2hi, line 387: I absolutely could not get this to work, once I got the first element from errors, errors was unhappy due to it being a move :(

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -344,12 +283,146 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                                debug!("Failed to set datapoint: {:?}", err);
+                                                for (id, error) in err.into_iter() {
+                                                    if let Some(metadata) = broker.get_metadata(id).await {
+                                                        let path = metadata.path.clone();
+                                                        let data_entry_error = convert_to_data_entry_error(&path, &error);
+                                                        errors.push(data_entry_error);
+                                                    }
+                                                }
+
+                                                if let Err(err) = sender.send(
```
wba2hi, line 384: right now we are sending out only errors, no success messages / acknowledges!

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -344,12 +283,146 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                            match &request.entry {
+                                                Some(entry) => match broker.get_id_by_path(&entry.path).await {
+                                                    Some(id) => {
+                                                        match validate_entry_update(&broker, &request, id).await {
+                                                            Ok(pair) => {
+                                                                updates.push((pair.first, pair.second));
+                                                            }
+                                                            Err(e) => {
+                                                                let message = format!("Data present in the request is invalid: {}", e.message());
+                                                                errors.push(proto::DataEntryError {
```
wba2hi, line 336: Some errors which previously early returning the function (see here: https://github.com/eclipse-kuksa/kuksa-databroker/pull/34/files#diff-764f0adb7e58cb82e14ca30770f38dce1b3a6d003d729ddd651442ece69e1509L235) were converted to DataEntryError
While for a synchronous set we want to fail fast, for the async streaming api we don't want to end the loop in case an error occurs due to an invalid data point.

SebastianSchildt:
@morgenroth this might fit your use case better than sdv1 API, and once goes to mainline, might stay around a bit longer

@wba2hi  afaik @morgenroth requires actuator streaming, so I guess the "field" needs to be set differently in your example?

wba2hi:
> @morgenroth this might fit your use case better than sdv1 API, and once goes to mainline, might stay around a bit longer
> 
> @wba2hi afaik @morgenroth requires actuator streaming, so I guess the "field" needs to be set differently in your example?

Yes, in case of setting an actuator the call would look like the following:
```
val datapoint = Datapoint.newBuilder()
    .setBool(true)
    .build()

val dataEntry = DataEntry.newBuilder()
    .setPath("Vehicle.Cabin.Door.Row1.DriverSide.IsOpen")
    .setValue(datapoint)

val entryUpdate = EntryUpdate.newBuilder()
    .addFields(Field.FIELD_ACTUATOR_TARGET)
    .setEntry(dataEntry)

val streamedUpdateRequest = StreamedUpdateRequest.newBuilder()
    .addUpdates(entryUpdate)
    .build()

streamingUpdate?.onNext(streamedUpdateRequest)
```

I can change it, but it's basically just a short proof of concept - the approach how it's implemented in the test app is not even good and not meant to be merged in this way, because we just start sending the request once we are connected to the databroker without manual interaction and also the delay of 1 second is also just introduced so we can see how the value changes between consecutive requests :D

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -344,6 +279,142 @@ impl proto::val_server::Val for broker::DataBroker {
...
+        };
+        let mut stream = request.into_inner();
+
+        let mut shutdown_trigger = self.get_shutdown_trigger();
+
+        // Copy (to move into task below)
+        let broker = self.clone();
+
+        // Create stream (to be returned)
+        let (sender, receiver) = mpsc::channel(10);
```
erikbosch, line 305: Should we possibly have a comment here explaining why we have selected the magic number 10. Why not 42?

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -344,12 +283,146 @@ impl proto::val_server::Val for broker::DataBroker {
...
+        // Listening on stream
+        tokio::spawn(async move {
+            let permissions = permissions;
+            let broker = broker.authorized_access(&permissions);
+            loop {
```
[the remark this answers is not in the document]
erikbosch: I am no Rust expert - anything that prevents dividing it into multiple methods to limit length and depth?

databroker/src/grpc/kuksa_val_v1/val.rs
wba2hi: I took it over from here: https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/databroker/src/grpc/sdv_databroker_v1/collector.rs#L106 

I have the feeling that value is set arbitrarily and can be set freely to 42 or even 100. The example code of channel goes with a "default" of 100. Shall I increase the value or just add a comment, that it's value has been set arbitrarily?

databroker/src/grpc/kuksa_val_v1/val.rs
lukasmittag: what you would do is do an 'errors.clone()' then you will not move the original errors array.

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -475,6 +546,95 @@ impl proto::val_server::Val for broker::DataBroker {
...
+    let update = broker::EntryUpdate::from_proto_entry_and_fields(entry, fields);
+
+    let pair = Pair {
+        first: id,
+        second: update,
+    };
+    Ok(pair)
+}
+
+fn convert_to_data_entry_error(path: &String, error: &broker::UpdateError) -> DataEntryError {
```
lukasmittag, line 585: would be nice to add a generic one for converting these maybe moving this into conversion.rs as well. https://github.com/eclipse-kuksa/kuksa-databroker/blob/82513eaf486dbc4e4e60fad0048b3cf8a10bfd8f/databroker/src/grpc/kuksa_val_v1/val.rs#L453 because here we do it in place.

databroker/src/grpc/kuksa_val_v1/val.rs
lukasmittag: Since it is a buffer https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html we should think about this. With this value after buffering 10 messages it will put any more messages on the channel and will wait.

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -15,20 +15,30 @@ use std::collections::HashMap;
...
 
 use crate::broker;
-use crate::broker::EntryReadAccess;
 use crate::broker::ReadError;
 use crate::broker::SubscriptionError;
+use crate::broker::{AuthorizedAccess, EntryReadAccess};
 use crate::glob;
 use crate::permissions::Permissions;
 
+struct Pair<F, S> {
```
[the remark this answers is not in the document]
wba2hi: It is used here to return the i32 id and EntryUpdate: https://github.com/eclipse-kuksa/kuksa-databroker/pull/34/files#diff-764f0adb7e58cb82e14ca30770f38dce1b3a6d003d729ddd651442ece69e1509R553

I didn't find a better name. In Java a Pair is a "normal" construct with two values (even though it is of course inferor to a well named type). Previously the stuff was just returning an anonymous (i32, EntryUpdate), however I could not reuse this for the Result. So I had to wrap it up.

proto/kuksa/val/v1/val.proto
```diff
@@ -37,6 +37,8 @@ service VAL {
   // Set entries
   rpc Set(SetRequest) returns (SetResponse);
 
+  rpc StreamedUpdate(stream StreamedUpdateRequest) returns (stream StreamedUpdateResponse);
```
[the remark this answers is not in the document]
wba2hi: From my memory it was decided to name it StreamedUpdate when we created the ticket (even though it was still unclear if now v1, v2 or whatever)

According to the ticket (https://etas-cloud.atlassian.net/browse/VAL-1320) and Databroker v2 API it will be named `provideStream` however I had the feeling that this does not fit very well for v1. In Databroker v2 API `set` is also named `provideOnce` therefore better fitting altogether. In the ticket it is also mentioning provideStream but I don't see this as a hard requirement.

However, I go with whatever you prefer.

databroker/src/grpc/kuksa_val_v1/val.rs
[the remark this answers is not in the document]
wba2hi: It probably makes sense to put this generic / util type somewhere better fitting instead of in val.rs

databroker/src/grpc/kuksa_val_v1/val.rs
wba2hi: These are custom errors, which are returned and not propagated back to the client using the DataEntryError, therefore I can not move them into the convert_to_data_entry_error method

databroker/src/grpc/kuksa_val_v1/val.rs
erikbosch: I assume the message size is not that big so having 100 is not a real problem, or? Anyway, whatever number we come up with I think it could be good to mention that it is an just an educated guess and not based on any real tests.

lukasmittag (dismissed):

databroker/src/grpc/kuksa_val_v1/val.rs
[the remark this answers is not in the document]
wba2hi: I could have sworn, that I tried it and it did not work with Result, however I tried again now and it works. So yes, I replaced it by an anonymous tuple:

```
async fn validate_entry_update(
    broker: &AuthorizedAccess<'_, '_>,
    request: &EntryUpdate,
    id: i32,
) -> Result<(i32, broker::EntryUpdate), Status>
```

proto/kuksa/val/v1/val.proto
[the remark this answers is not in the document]
rafaeling: I think we have come to a consensus on what to call it for V2, but for v1 I really don't know what to call it.
For sdv is  StreamDatapoints, for v2 ProvideStream, for v1 I also think it was StreamedUpdate but not really sure.

databroker/src/grpc/kuksa_val_v1/val.rs
argerus: I have a vague memory of measuring the throughput given different buffer sizes and that is how I ended up with a channel of size 10.

I guess a comment would have been a good thing to add then, but at the same time I don't think the testing was very thorough nor sound. But if we are to change it, we should probably do it while measuring the resulting behaviour.

erikbosch: Do we think this is a number that better should be a config (in the long term)? We previously had discussions on constants for max number of signals, max number of subscriptions and so on. Is something were think tweaking might be necessary for different deployment based on memory and traffic. I.e. in one system maybe 10 is a good choice as traffic never is high and memory scarce, in another it is the opposite. Or do we think we can have one number that fits all?

proto/kuksa/val/v1/val.proto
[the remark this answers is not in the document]
SebastianSchildt: No hard opinion, just my 2ct; I don"t think this will need to be written very often, so a "long" (and honestly it is not that long) name should be no issue. At least this is saying what it does sort of. Better than "setS" :D

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -743,4 +898,100 @@ mod tests {
...
+                    actuator_target: None,
+                }),
+            }],
+        };
+
+        let mut streaming_request = tonic_mock::streaming_request(vec![streamed_update_request]);
+        streaming_request
+            .extensions_mut()
+            .insert(permissions::ALLOW_ALL.clone());
+        match broker.streamed_update(streaming_request).await {
```
wba2hi, line 977: Calling broker.streamed_update requires a type Request<Streaming<StreamedUpdateRequest>>. However, after fidding around for hours I was not able to get this type without the help of the tonic_mock library. The tonic_mock library however forces us to update some lib versions (e.g. tonic, prost). Updating them introduces a lot of deprecation warnings and makes a lot of test fail: https://github.com/eclipse-kuksa/kuksa-databroker/actions/runs/9286289961/job/25552663996?pr=34

@argerus  Do you have an easy solution on how to create a "Request<Streaming<StreamedUpdateRequest>>"? 

Already tested stuff like this: 
- https://github.com/tyrchen/tonic-mock/pull/1#issuecomment-1241164173 (this basically should show an alternative solution on how we don't have to use tonic_mock but it might be to old by now?)
- https://www.qovery.com/blog/a-guided-tour-of-streams-in-rust/
- https://github.com/hyperium/tonic/blob/master/examples/routeguide-tutorial.md
- https://github.com/hyperium/tonic/blob/master/examples/src/streaming/server.rs#L97 (following the lines where it is called leads here: https://github.com/hyperium/tonic/blob/master/examples/src/streaming/client.rs#L11 but re-implementing this method for StreamedUpdateRequest does not work - it's simply not recognized by broker.streamed_update) 

None of them worked unfortunately and at this point in time I'm absolutely clueless what else to test

argerus: `tonic_mock::streaming_request(...)` seems to have been created specifically for making this task easier, correct?

If so, I would suggest using it. If I understand you correctly, the issue is then that `::from_i32(...)` has been deprecated in newer versions of tonic.

I pushed a [branch](https://github.com/boschglobal/kuksa-databroker/tree/topic/fix_deprecated_from_i32) based on this PR that changes the use of this deprecated function into using the `TryFrom` trait instead.

wba2hi: @argerus Thanks for your effort! I merged your branch into VAL-1320 and pushed the changes. 

Unfortunately that did not fix the failing tests. Maybe the issue comes from the cucumber setup but I am not able to debug it, making finding the root cause extra hard. I tried upgrading the version of cucumber to 0.21 but that did not fix it.

wba2hi: Nevermind, I found the error. It seems to no longer possible to connect to databroker via tls if not explicitely using "https://". Also "grpc://" is no longer supported. I think I saw the use of "grpc://" in some places, so adaptions to documentation might be necessary?

The following link helped me: https://github.com/hyper*ium/tonic/issues/1496

databroker/src/grpc/kuksa_val_v1/val.rs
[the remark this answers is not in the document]
wba2hi: By now it is possible to extract most of the loop into a separate method:
```
async fn handle_streamed_update_message(
    broker: &AuthorizedAccess,
    sender: &Sender<Result<StreamedUpdateResponse, Status>>,
    message: Result<Option<StreamedUpdateRequest>, Status>,
) {
```
However, here I'm facing some Rust specific issues again:
```
error[E0597]: `broker` does not live long enough
let broker = broker.authorized_access(&permissions);
argument requires that `broker` is borrowed for `'static`
```
and
```
error[E0597]: `permissions` does not live long enough
let broker = broker.authorized_access(&permissions);
argument requires that `permissions` is borrowed for `'static`
```

I don't easily want to make this change, because it would keep the instance alive as long as the application is running if I understood this correctly. 

So probably this can be fixed, however I don't know yet how the fix would be made appropriate.

databroker/tests/world/mod.rs
```diff
@@ -267,7 +267,7 @@ impl DataBrokerWorld {
 
         debug!("started Databroker [address: {addr}]");
 
-        let data_broker_url = format!("http://{}:{}", addr.ip(), addr.port());
+        let data_broker_url = format!("https://{}:{}", addr.ip(), addr.port());
```
SebastianSchildt, line 270: Why is this fixing the tests?

SebastianSchildt: Nevermind. Just seeing the comment now

databroker/src/grpc/kuksa_val_v1/val.rs
SebastianSchildt: That line in quickstart is unrelated, as it is using python code (grpc(s) an ws(s) scheme is a convention in the python CLI to quickly differnatiate  between  websocket based VISS or GRPC based kuksa protocol with or without TLS)

SebastianSchildt: If the tonic behavior changed, do we need some related change in rust  cli or does that anyway only ever forward URIs given by the user?

wba2hi: Well, good question, the databroker-cli as a default value uses `http://127.0.0.1:55555` (https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/databroker-cli/src/cli.rs#L40). 

However as a default the tls options are disabled. So you could argument and say it's fine, but we probably should document somewhere, that when using `https` the url needs to be explicitely specified using `https://`, maybe once in the corresponding tls section and maybe once in some error faq, where you can look up the error and see a fix for it?

wba2hi: I made some changes to the documentation with the latest commit and added some info about it, please have a look at it.

SebastianSchildt: Documentation update  good. Whether this PR techically "sound"/good enough I can not say

databroker/src/grpc/kuksa_val_v1/val.rs
SebastianSchildt: Probably a comment owuld be good. Whether such details need to be configurable: I am unsure, if this is a "user-serviceable setting" it would also imply more testing.

Anyway, lets not "solve" this in this PR

argerus (approved):
Looks good to me

PR #37, opened 2024-06-06
Push databroker container images also to quay.io

SebastianSchildt:
This will push databroker to quay.io. 

Can only really be tested once it is merged, as we can only see then whether the secrets work

See here for background https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4712

And I did a test with a private quay repo to check the "pattern" used here should work:

https://github.com/SebastianSchildt/testquay
https://github.com/SebastianSchildt/testquay/actions/runs/9386003500/job/25845517800

### DIFF
--- a/.github/workflows/kuksa_databroker_build.yml
+++ b/.github/workflows/kuksa_databroker_build.yml
@@ -178,6 +178,7 @@ jobs:
         # list of Docker images to use as base name for tags
         images: |
           ghcr.io/eclipse-kuksa/kuksa-databroker
+          quay.io/eclipse-kuksa/kuksa-databroker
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
After merge and successful  containers would be expected at 

```
docker pull quay.io/eclipse-kuksa/lkuksa-databroker:main
```

lukasmittag (approved):
LGTM

erikbosch:
Seems to work, see https://quay.io/repository/eclipse-kuksa/kuksa-databroker?tab=tags&tag=latest

But do we want to mention it in documentation as well, and if so just in a single place and keep ghcr.io as the default in all examples?

PR #38, opened 2024-06-06
Prepare Release 0.4.5

wba2hi:
Functional Check as per Release Process (https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Process#update-rust-versions) looked good. 

One thing I noticed when testing the `dbcfeeder.py`:
Instead of doing manual get's on Vehicle.Speed I subscribed to it using databroker-cli to be directly informed about changes. However, once I started `dbcfeeder.py` I got a lot of updates for "Vehicle.Speed: 0km/h", it took quite some time until it changed to actual numbers. This might be working as intended, however I don't know, therefore I want to highlight it

### DIFF
--- a/scripts/prepare_release.sh
+++ b/scripts/prepare_release.sh
@@ -32,7 +32,18 @@ SCRIPT_DIR=$(dirname "$SCRIPT_PATH")
 DATA_BROKER_ROOT=$SCRIPT_DIR/..
 
 # Update Cargo.toml versions.
-sed -i -E "s/^version = \"${VERSION_REGEX}\"$/version = \"${VERSION}\"/" \
+if [[ "$OSTYPE" == "darwin"* ]]; then
+	if ! command -v gsed &> /dev/null; then
+    	echo "gsed could not be found; install gnu-sed"
+    	exit 1
+	else
+		SED_COMMAND="gsed" # requires gnu-sed
+	fi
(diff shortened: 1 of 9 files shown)


scripts/prepare_release.sh
```diff
@@ -32,7 +32,18 @@ SCRIPT_DIR=$(dirname "$SCRIPT_PATH")
...
+    	echo "gsed could not be found; install gnu-sed"
+    	exit 1
+	else
+		SED_COMMAND="gsed" # requires gnu-sed
+	fi
+else
+	SED_COMMAND="sed"
+fi
+
+$SED_COMMAND -i -E "s/^version = \"${VERSION_REGEX}\"$/version = \"${VERSION}\"/" \
```
wba2hi, line 46: on Macs preinstalled version of sed `-i` expects an argument (backup file extension) thus making the script incompatible with Mac. Since it was not easily possible to somehow extract only the different handling of the `-i flag` we are now using the compatible gsed for Mac. 
Now it should work on both Linux and Mac

argerus (approved):
LGTM

SebastianSchildt (approved):
lgtm

wrt to DBC feeder: The observation is correct, the log was setup was started, when the car was in park, and it took some time before it moves

PR #40, opened 2024-06-06
Fix quay.io secrets forwarding

SebastianSchildt:

Issue #36, opened 2024-06-06
Quay.io onboarding - Container image accessibility


SebastianSchildt:
Making an issues, as this may take several PRs

Currently the KUKSA project pushes images (i.e. ghcr.io/eclipse-kuksa/kuksa-databroker:main)
This is an issue for our Chinese users as ghcr is blocked by the Great Firewall. Therfore, we intend to push images to quay.io as well. Quay is a Redhat (also Eclipse SDV member btw) operated Container registry, that is free for OSS projects AND that is accessible in China. The proposed solution is, in ADDITION to ghcr in our build pipelines pushing quay.io as well.

There is a related Eclipse Foundation ticket here: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4712

Steps

- [x]  Discuss with Eclipse Foundation how this can be set up and get necessary org secrets
- [x] Modify databroker build to push to quay as welll https://github.com/eclipse-kuksa/kuksa-databroker/pull/37
- [x] Manually push at least the 0.4 databroker releases to quay as well 
- [x] Modify databroker-cli build to push to quay as welll #41 
- [x] Manually push at least the 0.4 databroker-cli releases to quay as well
- [x] Update quickstart docs with a hint pointing to the alternative

SebastianSchildt:
Everything pushed Eclipse side. Keep this open, until all other components are modified to push to quay as well.

SebastianSchildt:
All done for now

PR #41, opened 2024-06-07
Push databroker-cli images also to quay.io

SebastianSchildt:
Does the same for clli ash #39 and #40 did for databroker

@wba2hi  Maybe it makes sense the retag 0.4.5 on final time after this in in, to make sure 
 - we did not mess up (unlikely, I know 😁 
 - Then we have it consistent on quay with 0.4.5 and don't have the situation where databroker is pushed by cli is not

(As per #36  I might still document this in a follow up PR, and also check if there are possibilities to upload older releases, but those are no reasons for blocking or re-tagging a release)

### DIFF
--- a/.github/workflows/create_draft_release.yml
+++ b/.github/workflows/create_draft_release.yml
@@ -58,6 +58,9 @@ jobs:
       QUAY_IO_USERNAME: ${{ secrets.QUAY_IO_USERNAME }}
   call_kuksa_databroker-cli_build:
     uses: ./.github/workflows/kuksa_databroker-cli_build.yml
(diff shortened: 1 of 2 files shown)


wba2hi:
> @wba2hi Maybe it makes sense the retag 0.4.5 on final time after this in in, to make sure
> 
>     * we did not mess up (unlikely, I know 😁
> 
>     * Then we have it consistent on quay with 0.4.5 and don't have the situation where databroker is pushed by cli is not

Yes, let's do it this way.

wba2hi (approved):
LGTM

PR #53, opened 2024-08-02
[Feature] Add features subscribe/publish_value v2

lukasmittag:
This adds the features of kuksa.val.v2 to subscribe and publish a value to databroker. It fixes some conversion functions as well. It adds a unit test for subscribe function that leverages `publish_value`.


lukasmittag:
false base branch

PR #58, opened 2024-09-04
Changing version on main to 0.4.7-dev.0 and update script

erikbosch:
Background:

If we get a log from someone it is currently impossible to say if they are using a released version of databroker or something thing have built themselves from `main`. By using a prerelease-version on main-branch we can assure that locally built Databrokers from main (unless they happen to use the exact commit used for a release) not will report a released version number like 0.4.6 or 0.4.7

This PR contain:

- Update of version to `0.4.7-pre.0`. if we believe next release from main will be `0.5.0` we could use `0.5.0-pre.0` instead.
- It updates the helper script to support pre-release and require patch level. It seems that we must specify patch level in Rust 8see https://crates.io/crates/semver), a version like `0.5`is not allowed in toml files.
- It also updates general dependencies. This one also covers #56 

I have tried to write down some thoughts on how we could use versioning in https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Process#proposed-versioning-scheme

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


lukasmittag (approved):
LGTM

erikbosch:
@mikehaller @argerus Feel free to take a look

PR #61, opened 2024-09-05
Append release name to create release artifacts.

erikbosch:
This change adds the release name identifier specified when triggering the draft release workflow.
Also add prefix to created release name (as that was used before)

Old style:

https://github.com/eclipse-kuksa/kuksa-databroker/releases/tag/0.4.6

image

New style:

Existing at https://github.com/erikbosch/kuksa-databroker/releases/tag/untagged-ba3e6e0b4a7e7e814be2

image

### DIFF
--- a/.github/workflows/create_draft_release.yml
+++ b/.github/workflows/create_draft_release.yml
@@ -83,7 +83,12 @@ jobs:
           pattern: databroker*
           merge-multiple: true
       - name: Display structure of downloaded files
-        run: ls -R build-artifacts
+        run: |
+             ls -R build-artifacts
+             cd build-artifacts
+             # Rename, add release name (usually tag)
+             for f in databroker*.tar.gz; do mv "$f" "$(echo "$f" | sed s/.tar.gz/-${{ needs.get_version.outputs.version }}.tar.gz/)"; done
+
 
       - name: Create release
         id: create_release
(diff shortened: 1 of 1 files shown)


erikbosch:
FYI @mikehaller - please check screenshot.

lukasmittag:
https://github.com/eclipse-kuksa/kuksa-databroker/actions/runs/10715619732 would this affect master tar files naming as well or will they stay the same?

erikbosch:
> https://github.com/eclipse-kuksa/kuksa-databroker/actions/runs/10715619732 would this affect master tar files naming as well or will they stay the same?

This change ONLY affects the release workflow. It does not affect

https://github.com/eclipse-kuksa/kuksa-databroker/actions/workflows/kuksa_databroker_build.yml
https://github.com/eclipse-kuksa/kuksa-databroker/actions/workflows/kuksa_databroker-cli_build.yml

They will still create artifacts without any label in file names. That sort of makes sense as those actions are triggered by quite many reasons, like pull request and merge where it is not obvious what label to use. 

The label/version used in this PR is the second argument given when triggering this workflow:

image

I.e. you first select branch or tag, then you select label to use for the release, which can be totally unrelated. In our [release process](https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Process#automated-release-workflow) we however recommend to first tag a commit and then use the same number in both fields. But it is totally possible if we need a release just for test purposes that we trigger it like:

image

(It will anyway be draft release only visible to committers until a committer decides to publish it)

SebastianSchildt (approved):
Seems fine to me, as the draft release workflow seems to be triggered only manually anyway, currently there seems to be a low chance this disturbs anything else

PR #62, opened 2024-09-09
Update dependencies

erikbosch:
Regenerated lock files to remove gix-path vulnerability


SebastianSchildt (approved):
Ah nice, before that was not possible as,vergen did not use newer gix. I think this was anyway only active during compile time. But updated deps always good :)

PR #65, opened 2024-09-11
Fix handling of viss-address parameter

### LINKED ISSUE #64: --viss-address is not used
Hello everyone,

during usage of the databroker with VISS support I noticed that there is a --viss-address command line argument that should set the address of the VISS server. However, this argument seems to be not functional right now.

When I check the code I noticed the following line where actually the value of --address is used, instead:
https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/databroker/src/main.rs#L435C51-L435C58

For my use case it's not a problem to change the address of the gRPC API also, but it is at least confusing that this argument has no effect.

SebastianSchildt:
This should fix #64 

Now the parameter has an effect, and bind adresses for GRPC API and VISS server can be different, in case you want to expose these services on different interfaces.

### DIFF
--- a/databroker/src/main.rs
+++ b/databroker/src/main.rs
@@ -298,12 +298,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 Arg::new("viss-address")
                     .display_order(31)
                     .long("viss-address")
-                    .help("VISS address")
+                    .help("Bind address for VISS server, if argument is not provided, the value of --address is used")
                     .action(ArgAction::Set)
                     .value_name("IP")
                     .required(false)
                     .env("KUKSA_DATABROKER_VISS_ADDR")
-                    .default_value("127.0.0.1"),
             )
             .arg(
                 Arg::new("viss-port")
@@ -429,10 +428,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     #[cfg(feature = "viss")]
     {
+        let viss_bind_addr = if args.contains_id("viss-address") {
+            args.get_one::<String>("viss-address").unwrap().parse()?
+        } else {
+            args.get_one::<String>("address").unwrap().parse()?
+        };
+
         let viss_port = args
             .get_one::<u16>("viss-port")
             .expect("port should be a number");
-        let viss_addr = std::net::SocketAddr::new(ip_addr, *viss_port);
+        let viss_addr = std::net::SocketAddr::new(viss_bind_addr, *viss_port);
 
         if args.get_flag("enable-viss") {
             let broker = broker.clone();


SebastianSchildt:
Anyone?

databroker/src/main.rs
```diff
@@ -429,10 +429,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     #[cfg(feature = "viss")]
     {
+        let viss_bind_addr = args.get_one::<String>("viss-address").unwrap().parse()?;
+
         let viss_port = args
             .get_one::<u16>("viss-port")
             .expect("port should be a number");
-        let viss_addr = std::net::SocketAddr::new(ip_addr, *viss_port);
+        let viss_addr = std::net::SocketAddr::new(viss_bind_addr, *viss_port);
```
lukasmittag, line 437: This should be same as ip_addr only port different?

SebastianSchildt: No, that was the bug before. There is a seperate command line parameter for the VISS bind adress, and it makes sense to  have that to bind VISS to another interface depending on the usecase. I.e. bind on a "southbound" Ethernet for GRPC based feeders and e.g. only binding "low(er) risk" VISS to e.g. an WiFi Interface serving mobile clients etc.

argerus: If `--viss-address` is not specified at startup but `--address` is, then that should be used also for the VISS interface, right?

Or in other words, only if `--viss-address` is explicitly specified should it be used.

SebastianSchildt: "Not sure", so currently it isn't and before --viss-address was ignored completely. I could change the behavior so that address is "for all" *unless* viss-adress overrides it.

However, using that logic wouldn't it be more logical to also have an addtional parameter grpc-(or val-) address that overrides it address only for GRPC? Or maybe let --adress being the "shortcut" for viss-address and val-address?

argerus: > Or maybe let --adress being the "shortcut" for viss-address and val-address?

I think that's what I'm suggesting?

The most common by far is to just use `--address`. That will set the address for both the GRPC and VISS service. Is that what you mean by shortcut?

And in order to support having those services listen to different addresses / interfaces, if `--viss-address` can be specified. That gives you full control to set both the address of GRPC service _and_ the address of the VISS service.

erikbosch (approved):
Looks good to me

Tested with

`cargo run --bin databroker --features viss -- --metadata data/vss-core/vss_release_4.0.json --insecure --enable-viss --viss-address 127.0.0.2`

and verifying with kuksa-client

`kuksa-client ws://127.0.0.2:8090`

PR #69, opened 2024-09-24
Add allowed values to metadata

argerus:


lukasmittag:
As quick fix okay. But shouldn't we add it in kuksa.val.v1/v2 as well as message in types?

rafaeling:
Should not be merged to https://github.com/boschglobal/kuksa-databroker/tree/prep-rel-0.5.0?

argerus:
> As quick fix okay. But shouldn't we add it in kuksa.val.v1/v2 as well as message in types?

This adds it to `kuksa.val.v1` and `sdv.databroker.v1`. `kuksa.val.v2` is not a thing (yet)

lukasmittag:
yes, in code but why only add `message Allowed` in types.proto of sdv.databroker.v1?

lukasmittag:
Okay, my fault. It is nested in `ValueRestriction` here https://github.com/eclipse-kuksa/kuksa-databroker/blob/7951c62702e3eb3052bcf8f4e2c3235cd5adb6f5/proto/kuksa/val/v1/types.proto#L127

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -792,7 +792,73 @@ fn proto_entry_from_entry_and_fields(
         }
         if all || fields.contains(&proto::Field::MetadataValueRestriction) {
             metadata_is_set = true;
-            // TODO: Add to Metadata
+            metadata.value_restriction = match entry.metadata().allowed.as_ref() {
+                Some(allowed) => match allowed {
+                    broker::DataValue::StringArray(vec) => Some(proto::ValueRestriction {
+                        r#type: Some(proto::value_restriction::Type::String(
```
erikbosch, line 798: As a Rust newbie, why do we need `r#type` here?

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -792,7 +792,73 @@ fn proto_entry_from_entry_and_fields(
...
+                            proto::ValueRestrictionString {
+                                allowed_values: vec.clone(),
+                            },
+                        )),
+                    }),
+                    broker::DataValue::Int32Array(vec) => Some(proto::ValueRestriction {
+                        r#type: Some(proto::value_restriction::Type::Signed(
+                            proto::ValueRestrictionInt {
+                                allowed_values: vec.iter().cloned().map(i64::from).collect(),
+                                min: None, // TODO: Implement
```
erikbosch, line 808: So for now just creating a placeholder for supporting min/max?

proto/sdv/databroker/v1/types.proto
```diff
@@ -143,16 +143,27 @@ message Datapoint {
...
-  // Id to be used in "get" and "subscribe" requests. Ids stay valid during
-  // one power cycle, only.
   int32 id               = 1;
   EntryType entry_type   = 2;
   string name            = 4;
   DataType data_type     = 5;
   ChangeType change_type = 6;  // CONTINUOUS or STATIC or ON_CHANGE
   string description     = 7;
 
+  Allowed allowed = 10;
```
erikbosch, line 166: Have we ever used the metadata on the lines below? Even if the lines below are commented I assume it would be better to not have the same number (10) on this line and the line below, or?

argerus: They have never been used. I guess it's better to just remove them.

databroker/src/grpc/kuksa_val_v1/val.rs
erikbosch: I have no strong opinion, depends on how much time that would take I think. If it is quick go for it, if not keep the TODOs and do it later.

proto/sdv/databroker/v1/types.proto
erikbosch: Then just remove them I would say

erikbosch (approved):
Looks good enough

PR #72, opened 2024-09-27
Remove oneof and ValueFailure from Datapoint and some proto naming improvements

rafaeling:


proto/kuksa/val/v2/types.proto
```diff
@@ -20,11 +20,7 @@ option go_package = "kuksa/val/v2";
 
 message Datapoint {
   google.protobuf.Timestamp timestamp = 1;
-
-  oneof value_state {
-    ValueFailure failure = 2;
-    Value value          = 3;
-  }
+  Value value                         = 2;
```
erikbosch, line 23: So what is now supposed to be returned if there is no value available, a `value` that is `None`?

Like if you do a get request, but no value available yet, should yet get a Datapoint back where value is None. If so, maybe we shall add a comment that value may be None to indicate no value available (but signal known)

BjoernAtBosch: I accept this change, but comment as proposed by Erik should be added.
Remark: 
If needed we could later re-add the oneof (with renamed fields) and still stay compatible with the current version:
```
message Datapoint {
  google.protobuf.Timestamp timestamp = 1;

  oneof value_state {
    Value value = 2;
    MissingReason missingValueReason = 3;
  }
}

enum MissingReason {
   ...
}
```

BjoernAtBosch (commented):
Proto files lgtm

proto/kuksa/val/v2/types.proto
erikbosch: @rafaeling - will you update so we can have it merged (If my interpretation of when None shall be used is correct)

rafaeling: > I accept this change, but comment as proposed by Erik should be added. Remark: If needed we could later re-add the oneof (with renamed fields) and still stay compatible with the current version:
> 
> ```
> message Datapoint {
>   google.protobuf.Timestamp timestamp = 1;
> 
>   oneof value_state {
>     Value value = 2;
>     MissingReason missingValueReason = 3;
>   }
> }
> 
> enum MissingReason {
>    ...
> }
> ```

I added it also to the discussion topic list - https://github.com/eclipse-kuksa/kuksa-databroker/pull/21/files#diff-7ed9bb015da6a2a5d42536ce062cb6df04a6b8a9688a4dcfe15995a3ef300087R556
maybe if we use one day the design documentation we could also add the discussion link why it was decided to design the proto message that way.

rafaeling: > @rafaeling - will you update so we can have it merged (If my interpretation of when None shall be used is correct)

Will update it

erikbosch (approved):
LGTM

PR #73, opened 2024-09-27
Implement Provider Actuation

wba2hi:


databroker/Cargo.toml
```diff
@@ -65,13 +65,14 @@ axum = { version = "0.6.20", optional = true, features = ["ws"] }
...
 uuid = { version = "1.4.1", optional = true, features = ["v4"] }
+async-trait = "0.1.82"
 
 # systemd related dependency, only relevant on linux systems
 [target.'cfg(target_os = "linux")'.dependencies]
 sd-notify = "0.4.1"
 
 [features]
-default = ["tls"]
+default = ["tls", "dep:futures"]
```
wba2hi, line 75: required for val.rs -> provide_actuation
```
    let future_vss_ids = vss_paths.iter().map(|vss_path| broker.get_id_by_path(&vss_path));
    let resolved_opt_vss_ids = futures::future::join_all(future_vss_ids).await;
```

databroker/src/broker.rs
```diff
@@ -179,6 +180,23 @@ pub struct DataBroker {
...
+}
+
+#[derive(Clone)]
+pub struct ActuationChange {
+    pub id: i32,
+    pub data_value: DataValue,
+}
+
+pub struct ActuationSubscription {
+    vss_ids: Vec<i32>,
```
wba2hi, line 195: vss_paths are normalized to vss_ids, so it's easier to keep a common list of available / registered actuators!

databroker/src/broker.rs
```diff
@@ -179,6 +180,23 @@ pub struct DataBroker {
...
+    actuation_provider: Box<dyn ActuationProvider + Send + Sync + 'static>,
+    permissions: Permissions,
+}
```
wba2hi, line 198: right now some cleanup operation is missing, e.g. an ActuationProvider disconnects or crashes and can no longer actuate. However, we prevent multiple ActuationProviders to register for the same signal, meaning once the ActuationProvider disconnects, it can no longer re-register to it's signal.

I can not use the actuation_provider to check if it is still available. I maybe could check (e.g. Err during send()) in val.rs -> Provider -> actuate but there I would not have the relation between the ActuationProvider and the ActuationSubscription. Looking them up (e.g. based on the vss_id) feels like a lot of overhead, not sure if there is a better way

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -482,18 +650,20 @@ impl proto::val_server::Val for broker::DataBroker {
...
-                                                break;
-                                            },
-                                            Some(PublishValuesRequest(publish_values_request)) => {
+                                           Some(PublishValuesRequest(publish_values_request)) => {
                                                 let response = publish_values(&broker, &publish_values_request).await;
                                                 if let Err(err) = response_stream_sender.send(Ok(response)).await
                                                 {
                                                     debug!("Failed to send response: {}", err);
                                                 }
+                                            },
```
wba2hi, line 659: Right now Errors are not properly propagated to the Provider - he won't know if registration fails (e.g. invalid signal) except of maybe due to the log. Since we are in a thread here, we can't return the err to the function body.

Should we even establish a connection in case the signal does not exist? (just to say 'Yeah, you got a channel now, but you will never receive any responses except of this one, telling you that you provided the wrong signal') Maybe some kind of pre-verification / pre-validation is needed.

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -482,18 +650,20 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                             Some(ProvideActuation(provided_actuation)) => {
+                                                let response = provide_actuation(&broker, &provided_actuation, response_stream_sender.clone()).await;
+                                                if let Err(err) = response_stream_sender.send(response).await
+                                                {
+                                                    debug!("Failed to provide actuation: {}", err)
+                                                }
+                                                break;
                                             },
                                             Some(BatchActuateStreamResponse(_batch_actuate_stream_response)) => {
                                                 if let Err(err) = response_stream_sender.send(Err(tonic::Status::new(tonic::Code::Unimplemented, "Unimplemented"))).await {
```
wba2hi, line 669: Handling here is unclear. What is expected to happen when a client (who is not an Actuation Provider) sends the DataBroker a BatchActuateStreamResponse?

databroker/src/broker.rs
```diff
@@ -179,6 +180,23 @@ pub struct DataBroker {
...
+#[derive(Clone)]
+pub struct ActuationChange {
+    pub id: i32,
+    pub data_value: DataValue,
+}
+
+pub struct ActuationSubscription {
+    vss_ids: Vec<i32>,
+    actuation_provider: Box<dyn ActuationProvider + Send + Sync + 'static>,
+    permissions: Permissions,
```
wba2hi, line 197: The permissions here are still unused: 
Where would be the best place to check the permissions? In val.rs / Provider / actuate (can't access them there though, since we only propagate the ActuationChanges and not the Subscriptions to val.rs)? The methods in broker itself already check permissions on call.

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -482,18 +650,20 @@ impl proto::val_server::Val for broker::DataBroker {
...
                                                 }
+                                            },
+                                             Some(ProvideActuation(provided_actuation)) => {
+                                                let response = provide_actuation(&broker, &provided_actuation, response_stream_sender.clone()).await;
+                                                if let Err(err) = response_stream_sender.send(response).await
+                                                {
+                                                    debug!("Failed to provide actuation: {}", err)
+                                                }
+                                                break;
                                             },
```
wba2hi, line 667: Right now Errors are not properly propagated to the Provider - he won't know if registration fails (e.g. invalid signal) except of maybe due to the log. Since we are in a thread here, we can't return the err to the function body.

Should we even establish a connection in case the signal does not exist? (just to say 'Yeah, you got a channel now, but you will never receive any responses except of this one, telling you that you provided the wrong signal') Maybe some kind of pre-verification / pre-validation is needed.

databroker/src/broker.rs
wba2hi: Checked before calling "actuation_provider#actuate" (see broker.rs#actuate method)

databroker/src/broker.rs
wba2hi: added is_available method to trait ActuationProvider. This is now checked in cleanup.

databroker/Cargo.toml
wba2hi: Made futures a mandatory / non-optional dependency

databroker/src/grpc/kuksa_val_v2/val.rs
wba2hi: as discussed: added a todo here

databroker/src/grpc/kuksa_val_v2/val.rs
wba2hi: In fact errors are now correctly propagated to the provider (see corresponding tests).

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -650,6 +898,23 @@ fn convert_to_proto_stream(
...
+    actuation_error: broker::ActuationError,
+    message: String,
+) -> tonic::Status {
+    match actuation_error {
+        broker::ActuationError::NotFound => tonic::Status::not_found(message),
+        broker::ActuationError::WrongType => tonic::Status::invalid_argument(message),
+        broker::ActuationError::OutOfBounds => tonic::Status::out_of_range(message),
+        broker::ActuationError::UnsupportedType => tonic::Status::invalid_argument(message),
+        broker::ActuationError::PermissionDenied => tonic::Status::permission_denied(message),
+        broker::ActuationError::PermissionExpired => tonic::Status::permission_denied(message),
```
erikbosch, line 911: @lukasmittag  - this is related to your ideas on what to return, permission denied or unauthenticated

lukasmittag: In my opinion just fine granularity but basically if the token has expired you're unauthenticated and need to re-authenticate if you get permission denied you do not know the cause but need to fetch and process the message as well. Unauthenticated can mean two things as well since you can be unauthenticated because you're token is not correct or the token has expired. So to conclude you would either way need to have a look at the message I guess.

lukasmittag (changes requested):

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -152,24 +210,148 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                }
+                None => {
+                    return Err(tonic::Status::new(
+                        tonic::Code::InvalidArgument,
+                        "No Signal_Id provided",
+                    ))
+                }
+            };
+        };
+        return Err(tonic::Status::invalid_argument(
```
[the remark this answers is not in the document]
wba2hi: Good catch! As there is no added value from using the longer variant (new + tonic::Code) we should use the shorter version imo.

databroker/src/grpc/kuksa_val_v2/val.rs
wba2hi: Changed PermissionExpired to return tonic::Status::unauthenticated. But basically you are right, the user has to read the message to know the exact issue :)

databroker/src/grpc/kuksa_val_v2/val.rs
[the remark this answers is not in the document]
wba2hi: I replaced all occurences of tonic::Status::new with the corresponding convenience method creations

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -650,6 +864,23 @@ fn convert_to_proto_stream(
     })
 }
 
+fn convert_actuation_error_to_status(
```
lukasmittag, line 867: please move this function to conversion.rs

databroker/src/broker.rs
```diff
@@ -1561,6 +1623,290 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+        let provided_vss_ids: Vec<i32> = self
+            .broker
+            .subscriptions
+            .read()
+            .await
+            .actuation_subscriptions
+            .iter()
+            .flat_map(|subscription| subscription.vss_ids.clone())
+            .collect();
+        let intersection: Vec<_> = vss_ids
```
lukasmittag, line 1679:
```suggestion
        let intersection: Vec<i32> = vss_ids
```
rust can identify that Vec<_> should be Vec<i32> but shouldn't we just specify that it is that? I find it more readable this way. Whats your opinion @argerus @rafaeling

databroker/src/broker.rs
```diff
@@ -1561,6 +1623,290 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+
+        Ok(())
+    }
+
+    async fn map_actuation_changes_by_vss_id(
+        &self,
+        actuation_changes: Vec<ActuationChange>,
+    ) -> HashMap<i32, Vec<ActuationChange>> {
+        let mut actuation_changes_per_vss_id: HashMap<i32, Vec<ActuationChange>> = HashMap::new();
+        for ele in actuation_changes {
```
lukasmittag, line 1710:
```suggestion
        for actuation_change in actuation_changes {
```

databroker/src/broker.rs
```diff
@@ -1561,6 +1623,290 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+            .add_actuation_subscription(actuation_subscription);
+
+        Ok(())
+    }
+
+    async fn map_actuation_changes_by_vss_id(
+        &self,
+        actuation_changes: Vec<ActuationChange>,
+    ) -> HashMap<i32, Vec<ActuationChange>> {
+        let mut actuation_changes_per_vss_id: HashMap<i32, Vec<ActuationChange>> = HashMap::new();
```
lukasmittag, line 1709:
```suggestion
        let mut actuation_changes_per_vss_id: HashMap<i32, Vec<ActuationChange>> = HashMap::HashMap::with_capacity(len(actuation_changes));
```
Since we know how big the HashMap could be only reserve the capacity we need.

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -146,30 +190,145 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                debug!(?permissions);
+                permissions.clone()
+            }
+            None => return Err(tonic::Status::unauthenticated("Unauthenticated")),
+        };
+        let broker = self.authorized_access(&permissions);
+        let actuate_requests = request.into_inner().actuate_requests;
+
+        let mut actuation_changes: Vec<ActuationChange> = vec![];
+        for ele in actuate_requests {
```
lukasmittag, line 297:
```suggestion
        for actuate_request in actuate_requests {
```

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -146,30 +190,145 @@ impl proto::val_server::Val for broker::DataBroker {
...
+    // Returns (GRPC error code):
+    //   NOT_FOUND if any of the actuators are non-existant.
+    //   PERMISSION_DENIED if access is denied for any of the actuators.
+    //   UNAVAILABLE if there is no provider currently providing an actuator
+    //   INVALID_ARGUMENT
+    //       - if the data type used in the request does not match
+    //            the data type of the addressed signal
+    //       - if the requested value is not accepted,
+    //            e.g. if sending an unsupported enum value
     async fn batch_actuate(
```
lukasmittag, line 281: If we would respond with some `BatchActuateResponse` in the future I think we need to split this as well into `batch_actuate_id` and `batch_actuate`. Otherwise the client could mix paths and ids and when getting back a response he could be confused. Or am I wrong @rafaeling @argerus ?

databroker/src/broker.rs
```diff
@@ -1561,6 +1623,290 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+        for actuation_change in &actuation_changes {
+            let vss_id = actuation_change.id;
+            let result_entry = self.get_entry_by_id(vss_id).await;
+            match result_entry {
+                Ok(entry) => {
+                    let vss_path = entry.metadata.path;
+                    let result_can_write_actuator =
+                        self.permissions.can_write_actuator_target(&vss_path);
+                    match result_can_write_actuator {
+                        Ok(_) => {}
```
lukasmittag, line 1747: Shouldn't we check here if a provider is available else we can break here no need to do two for loops. You can fill the map<vss_id, actuation_change> here as well. And then you just emit the actuation requests once you collected all actuation changes successfully

databroker/src/broker.rs
```diff
@@ -1561,6 +1623,290 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+                        .actuate(actuation_changes)
+                        .await?
+                }
+                None => {
+                    let message = format!("actuation provider for vss_id {} not available", vss_id);
+                    return Err((ActuationError::ProviderNotAvailable, message));
+                }
+            }
+        }
+
```
lukasmittag, line 1800: Here is the legacy code missing to fill `actuator_target`

databroker/src/broker.rs
```diff
@@ -1561,6 +1623,290 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+    ) -> Result<(), (ActuationError, String)> {
+        let vss_id = *vss_id;
+        let result_entry = self.get_entry_by_id(vss_id).await;
+        match result_entry {
+            Ok(entry) => {
+                let vss_path = entry.metadata.path;
+                let result_can_write_actuator =
+                    self.permissions.can_write_actuator_target(&vss_path);
+                match result_can_write_actuator {
+                    Ok(_) => {}
```
lukasmittag, line 1817: If you are here then you can directly hand the actuate request through or not?

databroker/src/broker.rs
```diff
@@ -1561,6 +1623,290 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+                let error = opt_error.unwrap();
+                let message = format!(
+                    "Could not set actuator target for vss_id {}: {:?}",
+                    vss_id, error.1
+                );
+                warn!(message);
+            }
+        }
+
+        let mut guard = self.broker.subscriptions.write().await;
```
lukasmittag, line 1876: you just need read acces here right? Like in 1731

lukasmittag (changes requested):

databroker/src/broker.rs
wba2hi: Would be possible. However I tried to not mix to many concerns here. Therefore this loop just cares about the permissions. 

Logically there also would be a minor difference in both approaches. If we handle the Actuation Request inside the ok it would mean, that we might send out three actuate request and on the forth we notice that an error occurs. This could leave the database in an inconsistent state (?) because some requests are executed and some aren't. Right now the Actuation Requests are only forwarded once all permissions etc are verified.

databroker/src/broker.rs
wba2hi: I execute cleanup in line 1890 after checking if any actuation provider expired, so I think right now it's still feasible. However it's still challengable if I should call cleanup here, or we maybe don't need to call it at all.

databroker/src/broker.rs
wba2hi: Would be possible. However I tried to not mix to many concerns here. Therefore this loop just cares about the permissions.

Having multiple loops shouldn't be the issue however having to do multiple id lookups might be bad performance-wise. Since the creation of the Map<vss_id, Vec<ActuationChange>> is a complex method on it's own I would prefer leaving it in it's own method instead of inlining the code inside of the ok block. Here I vote for readability / maintainability over performance.

However, if you have a good idea how to achieve both, please feel free to make a short pseudo-code suggestion (or let's have another call).

databroker/src/broker.rs
wba2hi: Good catch, added!

databroker/src/broker.rs
wba2hi: While trying to introduce the cleanup operation in batch_actuate it caused a panic during `let legacy_result = self.update_entries(entry_updates).await;`. The panic must somehow be caused by the write guard, be. So I removed the cleanup from both methods again and converted the guard back to a read guard. 

As the housekeeping task is executed every second, it should be sufficient to not require manually triggering the cleanup. 

However I also added the expiration check, which was missing from batch_actuate.

databroker/src/broker.rs
```diff
@@ -679,6 +719,28 @@ impl Subscriptions {
...
+                false
+            } else {
+                match sub.permissions.expired() {
+                    Ok(()) => true,
+                    Err(PermissionError::Expired) => {
+                        info!("Token expired for actuation provider: removing subscription");
+                        false
+                    }
+                    Err(err) => {
+                        info!(
```
erikbosch, line 735: This is not expected, right? As expired() only can return Ok and Expired right?

Change to `warn!` and included "unexpected" in text or comment?

databroker/src/broker.rs
```diff
@@ -679,6 +719,28 @@ impl Subscriptions {
                 }
             }
         });
+
+        self.actuation_subscriptions.retain(|sub| {
+            if !sub.actuation_provider.is_available() {
+                info!("actuation provider gone: removing subscription");
+                false
+            } else {
+                match sub.permissions.expired() {
```
erikbosch, line 728: I think this method has a somewhat strange name. Would it not be better to rename it to `is_valid()`, `valid()` or similar. Right now just by looking athe name I get the impression that it will return True/ok if it actually has expired.

databroker/src/broker.rs
```diff
@@ -1530,6 +1592,278 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+
+                            actuation_subscription
+                                .actuation_provider
+                                .actuate(actuation_changes)
+                                .await?
+                        }
+                    }
+                }
+                None => {
+                    let message = format!("actuation provider for vss_id {} not available", vss_id);
```
erikbosch, line 1778: In most other error message we start with capital letter, ie. `"Actuation provider ..."`.

databroker/src/broker.rs
```diff
@@ -1530,6 +1592,278 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+                    }
+                    Err(PermissionError::Expired) => {
+                        return Err((
+                            ActuationError::PermissionExpired,
+                            "Permission expired".to_string(),
+                        ))
+                    }
+                }
+            }
+            Err(ReadError::NotFound) => {
```
erikbosch, line 1813: This is the third time we have the code snippet below. I have not checked the code structure that thoroughly, but it would be possible and feasible to avoid code duplication by using a helper method?

```
Err(ReadError::NotFound) => {
                    let message = format!("Could not resolve vss_path of vss_id {}", vss_id);
                    return Err((ActuationError::NotFound, message));
                }
```

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -27,44 +28,92 @@ use databroker_proto::kuksa::val::v2::{
...
+            );
+
+        let response = OpenProviderStreamResponse {
+            action: Some(batch_actuate_stream_request),
+        };
+
+        let result = self.sender.send(Ok(response)).await;
+        if result.is_err() {
+            let send_error = result.unwrap_err().0;
+            if send_error.is_err() {
```
erikbosch, line 77: This if-statement is quite useless as we anyway return an error after, or?

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -146,30 +190,139 @@ impl proto::val_server::Val for broker::DataBroker {
...
-            "Unimplemented",
-        ))
+        Err(tonic::Status::unimplemented("Unimplemented"))
     }
 
+    // Actuate a single actuator
+    //
+    // Returns (GRPC error code):
+    //   NOT_FOUND if the actuator does not exist.
+    //   PERMISSION_DENIED if access is denied for of the actuator.
```
erikbosch, line 200: Typo, `for of`

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -535,10 +681,75 @@ impl proto::val_server::Val for broker::DataBroker {
...
+
+    let future_vss_ids = vss_paths
+        .iter()
+        .map(|vss_path| broker.get_id_by_path(vss_path));
+    let resolved_opt_vss_ids = futures::future::join_all(future_vss_ids).await;
+
+    for (index, opt_vss_id) in resolved_opt_vss_ids.iter().enumerate() {
+        if opt_vss_id.is_none() {
+            let message = format!(
+                "could not resolve id of vss_path: {}",
```
erikbosch, line 710: Capital C, i.e. `Could not ..."`

databroker/src/grpc/kuksa_val_v2/val.rs
wba2hi: Yes, you are right

databroker/src/broker.rs
wba2hi: Changed to a more generic: 
```
match &sub.permissions.expired() {
    Ok(()) => true,
    Err(permission_error) => {
        info!("Removing Provider due to {:?}", permission_error);
        false
    }
}
```

databroker/src/broker.rs
wba2hi: Yes, I had the same feeling while using the method. I think the method was written like this for easy reuse / easy return in the other methods e.g. permissions.rs#can_write_actuator_target()
```
    pub fn can_write_actuator_target(&self, path: &str) -> Result<(), PermissionError> {
        self.expired()?;

        if self.actuate.is_match(path) {
            return Ok(());
        }
        Err(PermissionError::Denied)
    }
```

The easiest way would probably be to simply return a bool in #expired() and simply adapt the other methods to a minimal more verbose version:
```
    pub fn can_write_actuator_target(&self, path: &str) -> Result<(), PermissionError> {
        if self.expired() {
            return Err(PermissionError::Expired);
        }

        if self.actuate.is_match(path) {
            return Ok(());
        }
        Err(PermissionError::Denied)
    }
```

databroker/src/broker.rs
wba2hi: Could you elaborate here? 
The ActuationError-specific error messages make it kinda hard to separate this into a separate method. I could add a ReadError -> ActuationError conversion (same obviously for a PermissionError -> ActuationError), but I would still need to provide the error message as a parameter so it contains useful data like vss_id or vss_path. Therefore I don't see a lot of gain.

erikbosch: We have calls to             `self.get_entry_by_id(vss_id).await;` on line 1585, 1689 and 1772, i.e. in all 3 actuate methods, so it was just a thought if we could reuse that part by adding the call and the common error handling in a helper method. Something like

```
result = self.MyActuateHelperMethod
match result {
  Ok(res) => ... // Do something
  Err(err)  => return Err(err), // Reuse the error created by helper method
}

```

wba2hi: Good point, done! 
Encapsulated the logic into a separate "can_write_actuator_target" method, which takes a vss_id, looks it up and checks if permissions to write them are available. Really saved a few lines of code.

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -220,30 +265,139 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                            return Err(tonic::Status::invalid_argument(
+                                "Signal needs to provide Path or Id",
+                            ))
+                        }
+                    };
+                }
+                None => return Err(tonic::Status::invalid_argument("No Signal_Id provided")),
+            };
+        };
+        return Err(tonic::Status::invalid_argument(
```
argerus, line 332: This is totally optional, but if you want less nesting there's quite a few places where you can replace a `match` with a combination of `ok_or(err)` and `?`.
```suggestion
                debug!(?request);
        let permissions = request
            .extensions()
            .get::<Permissions>()
            .ok_or(tonic::Status::unauthenticated("Unauthenticated"))?
            .clone();
        debug!(?permissions);
        let broker = self.authorized_access(&permissions);

        let actuator_request = request.into_inner();
        let value = actuator_request
            .value
            .ok_or(tonic::Status::invalid_argument(
                "Invalid Actuator Request provided",
            ))?;

        let signal = actuator_request
            .signal_id
            .ok_or(tonic::Status::invalid_argument("No signal_id provided"))?
            .signal;

        match &signal {
            Some(proto::signal_id::Signal::Path(path)) => {
                let id = broker
                    .get_id_by_path(path)
                    .await
                    .ok_or(tonic::Status::not_found(format!(
                        "Invalid vss_path provided {}",
                        path
                    )))?;

                match broker.actuate(&id, &DataValue::from(value)).await {
                    Ok(()) => Ok(tonic::Response::new(ActuateResponse {})),
                    Err(error) => Err(error.0.to_tonic_status(error.1)),
                }
            }
            Some(proto::signal_id::Signal::Id(id)) => {
                match broker.actuate(id, &DataValue::from(value)).await {
                    Ok(()) => Ok(tonic::Response::new(ActuateResponse {})),
                    Err(error) => Err(error.0.to_tonic_status(error.1)),
                }
            }
            None => Err(tonic::Status::invalid_argument(
                "Signal needs to provide Path or Id",
            )),
        }
```

wba2hi: Thanks for the suggestion! As I'm still a newbie with Rust I'm in particular interested in these kinda "shortcuts" which make code look better and reduce complexity and nesting - so if you have further just keep 'em coming :) I also had the feeling that all of these match-cascades couldn't be the only way to handle it, but I just didn't knew better

.github/workflows/kuksa_databroker-cli_build.yml
```diff
@@ -72,7 +72,7 @@ jobs:
 
   build:
     name: Build databroker-cli
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
wba2hi, line 75: Build failed suddenly due to the use of ubuntu-latest with the following message: 

```
Run cargo install cross cargo-license cargo-cyclonedx
    Updating crates.io index
     Ignored package `cross v0.2.5` is already installed, use --force to override
    Updating crates.io index
    Updating crates.io index
     Ignored package `cargo-cyclonedx v0.5.5` is already installed, use --force to override
     Summary Successfully installed cross, cargo-license, cargo-cyclonedx!
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Error: Process completed with exit code 1.
```

The following comment probably describes the problem best: https://git hub.com/PyO3/maturin-action/issues/291#issue-2564997789

We can either add --break-system-packages or downgrade the image back to 22.04 - I decided to pin the version to ubuntu-22.04 for the build step as an intermediate step to resolve the issue.

wba2hi:
Rebased and squashed the branch to a single commit

erikbosch (approved):
LGTM

argerus (approved):

PR #74, opened 2024-09-28
Sri hari patch 1

Sri-hari:


argerus:
We already generate a SBOM, so I don't see the need for this PR

Futhermore, it's not clear why it should be pushed to 34.46.105.78 with plain text credentials in the workflow file.

Closing it. Create an issue, if there is some unknown to me reason for this PR.

PR #75, opened 2024-09-30
Add GetValue and GetValues Support

erikbosch:


databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -45,7 +45,10 @@ impl From<&proto::Datapoint> for broker::Datapoint {
 impl From<broker::Datapoint> for Option<proto::Datapoint> {
     fn from(from: broker::Datapoint) -> Self {
         match from.value {
-            broker::DataValue::NotAvailable => None,
+            broker::DataValue::NotAvailable => Some(proto::Datapoint {
```
erikbosch, line 48: This one is worth discussing. Previously (for V2) we  just ignored "NotAvailable", i.e. it was never returned in subscribe. But at lest for GetValue we want to return it, so I move the filtering higher up.

But how do we want it to work for Subscribe? If we intend to support actively removing a DataPoint by providing a DataPoint with value Null, then maybe we want the subscribe to also have support for returning Null.

And if so, would it not make sense to also return Null in the first request when subscription is started?

(Currently PR does not change behavior for Subscribe)

erikbosch: Discussed among committers - it seems we want None to be returned also for subscribe. I will check how it affects test case and update the PR

erikbosch: It returns `None` there, but as it returns `None` there it means that the value will be discarded (and not sent to client) at https://github.com/eclipse-kuksa/kuksa-databroker/blob/feature/databroker-api-v2/databroker/src/grpc/kuksa_val_v2/val.rs#L638

Thats why we need to create a Datapoint with value `None` here.

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -39,12 +39,42 @@ const MAX_REQUEST_PATH_LENGTH: usize = 1000;
...
+            Ok(signal_id) => signal_id,
+            Err(err) => return Err(err),
+        };
+
+        let datapoint = match broker.get_datapoint(signal_id).await {
+            Ok(datapoint) => datapoint,
+            Err(ReadError::NotFound) => {
+                return Err(tonic::Status::new(tonic::Code::NotFound, "Path not found"))
+            }
+            Err(ReadError::PermissionDenied | ReadError::PermissionExpired) => {
```
lukasmittag, line 67: shouldn't we return UNAUTHENTICATED if the permissions are expired?

proto/kuksa/val/v2/val.proto
```diff
@@ -21,33 +21,45 @@ import "kuksa/val/v2/types.proto";
...
 
   // Get the latest values of a set of signals.
   // The returned list of data points has the same order as the list of the request.
+  // If a requested signal has no value a DataPoint where value is None will be returned.
   //
   // Returns (GRPC error code):
   //   NOT_FOUND if any of the requested signals doesn't exist.
   //   PERMISSION_DENIED if access is denied for any of the requested signals.
+  //
   rpc GetValues(GetValuesRequest) returns (GetValuesResponse);
```
lukasmittag, line 42: General question: Do we need to split this as well by `GetValuesId' and `GetValues`?

databroker/src/grpc/kuksa_val_v2/val.rs
erikbosch: I think we have been a bit inconsistent with that in the past, but I think it makes sense. Any opinion @argerus?

proto/kuksa/val/v2/val.proto
erikbosch: The message `SignalID` supports both, so you can use whatever you like with the same RPC call.

databroker/src/grpc/kuksa_val_v2/val.rs
erikbosch: Changed it in latest commit

proto/kuksa/val/v2/val.proto
```diff
@@ -130,15 +142,15 @@ message GetValuesRequest {
 }
 
 message GetValuesResponse {
-  repeated Datapoint datapoints = 1;
+  repeated Datapoint data_points = 1;
```
erikbosch, line 145: Renaming as we use `data_point` elsewhere so trying to be consistent.

databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -45,7 +45,10 @@ impl From<&proto::Datapoint> for broker::Datapoint {
...
+                value: None,
+                timestamp: None,
```
lukasmittag, line 50: why should it not have a timestamp?

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -39,22 +39,96 @@ const MAX_REQUEST_PATH_LENGTH: usize = 1000;
...
+        };
+
+        let datapoint = match broker.get_datapoint(signal_id).await {
+            Ok(datapoint) => datapoint,
+            Err(ReadError::NotFound) => return Err(tonic::Status::not_found("Path not found")),
+            Err(ReadError::PermissionDenied) => {
+                return Err(tonic::Status::permission_denied("Permission denied"))
+            }
+            Err(ReadError::PermissionExpired) => {
+                return Err(tonic::Status::unauthenticated("Unauthenticated"))
```
lukasmittag, line 69: Wouldn't it be beneficial if we tell them in the message "Unauthenticated because your token expired"? 

```
 Err(ReadError::PermissionExpired) => {
                    return Err(tonic::Status::unauthenticated(format!(
                        "Permission expired (id: {})",
                        signal_id
                    )))
```
like here in line 120 ff.

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -39,22 +39,96 @@ const MAX_REQUEST_PATH_LENGTH: usize = 1000;
...
+                debug!(?permissions);
+                permissions.clone()
+            }
+            None => return Err(tonic::Status::unauthenticated("Unauthenticated")),
+        };
+
+        let broker = self.authorized_access(&permissions);
+
+        let requested = request.into_inner().signal_ids;
+        let mut response_datapoints = Vec::new();
```
lukasmittag, line 94: shouldn't we have a map here instead. Since it's async the order is not guaranteed to be the same as requested. Maybe a general discussion point as well @rafaeling @argerus

argerus: Why would the order be affected by whether it's async? Or exactly what do you mean by "Since it's async"?

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -660,43 +734,679 @@ mod tests {
...
         }
     }
 
+    #[tokio::test]
+    async fn test_get_values_id_one_signal_ok() {
+        let config = GetValuesConfigBuilder::new()
+            .first_exist()
+            .request_first()
+            .send_auth()
+            .auth_first()
```
rafaeling, line 1260: If `auth_first` or `auth_second` are set to `True` should not be `send_auth` `True` by default?

rafaeling (commented):
Looks good to me, thanks for the documenting the proto files and thanks a lot for adding those really nice unit tests :)

databroker/src/grpc/kuksa_val_v2/conversions.rs
erikbosch: Fo "normal" DataPoints that has a value the timestamp represents the time the databroker registered that value (doesn't it?). In this case Databroker "deep level magic" has no value and provide no timestamp (I think). We could opt for adding current time if we like to show that it was None at Now, but "Now" is the only time we have.

But a good point to dicuss, do we want a timestamp when return Value==None and if so what Timestamp. (Theoretically we could request Databroker to keep track of when a Value change from something to None, but I do not think it do it at the moment)

databroker/src/grpc/kuksa_val_v2/val.rs
erikbosch: Makes sens to say that auth expired. 

There are theoretically a lot of GRPC error codes (see for example https://grpc.github.io/grpc/cpp/md_doc_statuscodes.html) that we cannot control but the client still can receive. But it might be a good policy to mention all that we explicitly return in our implementation

databroker/src/grpc/kuksa_val_v2/conversions.rs
erikbosch: It actually seems to return something, at least in some cases, so I will see if I can use it

erikbosch: So it returns registration time for the signal for "virgin" signals, we could use that.

databroker/src/grpc/kuksa_val_v2/val.rs
erikbosch: In the Proto file we say that the result shall be returned in the same order as requested and here it gets the values one by one so should work. I have some test cases which at least check that it work for two signals (Which however does not prove anything)

databroker/src/grpc/kuksa_val_v2/val.rs
erikbosch: Well, at least there is no reason to have auth_first as True if you do not intend to send it. There are generally four combos I check:

- We do not send credentials at all
- We send credentials but they are empty
- We send credentials for signal 1
- We send credentials for signal 2
- We send credentials for signal 1 and 2

I can change so that `send_auth` becomes optional if you call one of the other

proto/kuksa/val/v2/val.proto
```diff
@@ -21,33 +21,49 @@ import "kuksa/val/v2/types.proto";
...
 service VAL {
   // Get the latest value of a signal
+  // If the signal exist but does not have a valid value
+  // a DataPoint where value is None shall be returned.
   //
   // Returns (GRPC error code):
   //   NOT_FOUND if the requested signal doesn't exist
   //   PERMISSION_DENIED if access is denied
+  //
+  //
```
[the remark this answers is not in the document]
erikbosch: Have you not read the new Proto guidelines?

( No I am just kidding, will remove one)

databroker/src/grpc/kuksa_val_v2/conversions.rs
lukasmittag: yes, that's what I thought of and IF the signal provider would go away you would have a timestamp as well.

databroker/src/grpc/kuksa_val_v2/val.rs
lukasmittag: Yeah you're right I've missed the await call all good will resolve this conversation :)

databroker/src/grpc/kuksa_val_v2/conversions.rs
argerus: > So it returns registration time for the signal for "virgin" signals, we could use that.

Setting the value to `NotAvailable` seems to behave just like every other value, at least when it comes to `sdv.databroker.v1` (and hence it's supported by the core of databroker).

* Setting a value to `20.0` will record the timestamp om when that happened.
* Setting a value to `NotAvailable` will record the timestamp of when that happened.

databroker/src/grpc/kuksa_val_v2/val.rs
rafaeling: Maybe then leave it as it is, is not a hurdle to call `send_auth` if `auth_first` or `auth_second` are set to `True`

rafaeling: You changed it already, thanks

wba2hi (approved):
Tested using the following commands: 
> ghz --insecure --skipFirst 100 --total 100100 --proto ./proto/kuksa/val/v2/val.proto --call kuksa.val.v2.VAL.GetValue -d '{"signal_id": {"path": "Vehicle.Speed"}}' 0.0.0.0:55556
```
Summary:
  Count:	100000
  Total:	8.96 s
  Slowest:	17.98 ms
  Fastest:	0.18 ms
  Average:	4.20 ms
  Requests/sec:	11159.41

Response time histogram:
  0.177  [1]     |
  1.958  [4235]  |∎∎∎
  3.738  [25572] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  5.519  [57218] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  7.299  [11867] |∎∎∎∎∎∎∎∎
  9.080  [922]   |∎
  10.860 [93]    |
  12.641 [52]    |
  14.421 [37]    |
  16.202 [1]     |
  17.982 [2]     |

Latency distribution:
  10 % in 2.50 ms
  25 % in 3.53 ms
  50 % in 4.25 ms
  75 % in 4.72 ms
  90 % in 5.84 ms
  95 % in 6.51 ms
  99 % in 7.36 ms

Status code distribution:
  [OK]   100000 responses
```

ghz --insecure --skipFirst 100 --total 100100 --proto ./proto/kuksa/val/v2/val.proto --call kuksa.val.v2.VAL.GetValue -d '{"signal_id": {"id": 100}}' 0.0.0.0:55556
```
Summary:
  Count:	100000
  Total:	8.68 s
  Slowest:	12.33 ms
  Fastest:	0.15 ms
  Average:	4.07 ms
  Requests/sec:	11525.19

Response time histogram:
  0.151  [1]     |
  1.368  [1093]  |∎
  2.586  [10441] |∎∎∎∎∎∎∎∎
  3.803  [23750] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  5.021  [49185] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  6.238  [10316] |∎∎∎∎∎∎∎∎
  7.456  [4902]  |∎∎∎∎
  8.673  [203]   |
  9.891  [52]    |
  11.108 [23]    |
  12.326 [34]    |

Latency distribution:
  10 % in 2.46 ms
  25 % in 3.42 ms
  50 % in 4.14 ms
  75 % in 4.59 ms
  90 % in 5.58 ms
  95 % in 6.27 ms
  99 % in 7.02 ms

Status code distribution:
  [OK]   100000 responses
```

ghz --insecure --skipFirst 100 --total 100100 --proto ./proto/kuksa/val/v2/val.proto --call kuksa.val.v2.VAL.GetValues -d '{"signal_ids": [{"path": "Vehicle.Speed"},{"path": "Vehicle.ADAS.ABS.IsEnabled"},{"id": 100}]}' 0.0.0.0:55556

```
Summary:
  Count:	100000
  Total:	8.90 s
  Slowest:	19.57 ms
  Fastest:	0.19 ms
  Average:	4.18 ms
  Requests/sec:	11241.01

Response time histogram:
  0.194  [1]     |
  2.131  [5680]  |∎∎∎∎
  4.068  [34113] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  6.005  [52709] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  7.942  [7293]  |∎∎∎∎∎∎
  9.880  [111]   |
  11.817 [43]    |
  13.754 [0]     |
  15.691 [0]     |
  17.629 [4]     |
  19.566 [46]    |

Latency distribution:
  10 % in 2.54 ms
  25 % in 3.58 ms
  50 % in 4.25 ms
  75 % in 4.69 ms
  90 % in 5.68 ms
  95 % in 6.37 ms
  99 % in 7.14 ms

Status code distribution:
  [OK]   100000 responses
```
Also tested the error case, where invalid VSS Paths or invalid ids are used. 
Performance-wise it does not look much worse than compared with kuksa.val.v1.Get: 

ghz --insecure --skipFirst 100 --total 100100 --proto ./proto/kuksa/val/v1/val.proto --call kuksa.val.v1.VAL.Get -d '{"entries": {"path": "Vehicle.Speed", "view": 1, "fields": 2}}' 0.0.0.0:55556

```
Summary:
  Count:	100000
  Total:	9.97 s
  Slowest:	16.46 ms
  Fastest:	0.52 ms
  Average:	4.72 ms
  Requests/sec:	10030.76

Response time histogram:
  0.519  [1]     |
  2.113  [1758]  |∎
  3.706  [13936] |∎∎∎∎∎∎∎∎∎
  5.300  [60628] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  6.894  [18435] |∎∎∎∎∎∎∎∎∎∎∎∎
  8.488  [4938]  |∎∎∎
  10.081 [242]   |
  11.675 [51]    |
  13.269 [10]    |
  14.863 [0]     |
  16.457 [1]     |

Latency distribution:
  10 % in 3.19 ms
  25 % in 4.15 ms
  50 % in 4.72 ms
  75 % in 5.26 ms
  90 % in 6.22 ms
  95 % in 6.93 ms
  99 % in 7.86 ms

Status code distribution:
  [OK]   100000 responses
```

Behavior is correct, therefore LGTM

erikbosch:
Thanks for testing @wba2hi

PR #76, opened 2024-10-01
Move conversion of UpdatError

lukasmittag:
UpdateError conversion for the new kuksa.val.v2 interface was done in the wrong place e.g. broker.rs. Move it now to conversion.rs.


rafaeling (approved):
LGTM

PR #77, opened 2024-10-02
Add metadata v2 value restriction

rafaeling:


databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -207,6 +207,99 @@ impl From<&proto::Datapoint> for broker::DataValue {
...
+            data_type: proto::DataType::from(metadata.data_type.clone()) as i32,
+            entry_type: proto::EntryType::from(metadata.entry_type.clone()) as i32,
+            description: Some(metadata.description.clone()),
+            comment: None,
+            deprecation: None,
+            unit: metadata.unit.clone(),
+            value_restriction: match metadata.allowed.as_ref() {
+                Some(allowed) => match allowed {
+                    broker::DataValue::StringArray(vec) => Some(proto::ValueRestriction {
+                        r#type: Some(proto::value_restriction::Type::String(
```
erikbosch, line 223: As a Rust newbie - why do we need `r#type` here?

databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -207,6 +207,99 @@ impl From<&proto::Datapoint> for broker::DataValue {
...
+                                },
+                            )),
+                        })
+                    }
+                    broker::DataValue::Int64Array(vec) => {
+                        Some(proto::ValueRestriction {
+                            r#type: Some(proto::value_restriction::Type::Signed(
+                                proto::ValueRestrictionInt {
+                                    allowed_values: vec.clone(),
+                                    min: None, // TODO: Implement
```
erikbosch, line 245: To be fixed before or after merge?

erikbosch: Noticed this is inherited from #69

databroker/src/grpc/kuksa_val_v2/conversions.rs
rafaeling: This is because of protobuf generator to avoid conflicts. It generates `r#type` if a field's name is declared as `type`.

erikbosch: Ah, `type` is possibly a reserved word in Rust?

erikbosch (approved):
LGTM and Ready for Review, so merging

PR #79, opened 2024-10-15
Min/Max first increment

erikbosch:


databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -792,72 +793,159 @@ fn proto_entry_from_entry_and_fields(
...
+            match entry.metadata().data_type {
+                DataType::String | DataType::StringArray => {
+                    let allowed = match entry.metadata().allowed.as_ref() {
+                        Some(broker::DataValue::StringArray(vec)) => vec.clone(),
+                        _ => Vec::new(),
+                    };
+
+                    if !allowed.is_empty() {
+                        metadata.value_restriction = Some(proto::ValueRestriction {
+                            r#type: Some(proto::value_restriction::Type::String(
```
lukasmittag, line 806: Is this something generated r#type or a typo?

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -792,72 +793,159 @@ fn proto_entry_from_entry_and_fields(
...
+                            }
+                            broker::DataValue::DoubleArray(vec) => vec.to_vec(),
+                            _ => Vec::new(),
+                        },
+                        _ => Vec::new(),
+                    };
+
+                    if min_value.is_some() | max_value.is_some() | !allowed.is_empty() {
+                        metadata.value_restriction = Some(proto::ValueRestriction {
+                            r#type: Some(proto::value_restriction::Type::FloatingPoint(
```
lukasmittag, line 935: since its here too I guess intended

lukasmittag (commented):
should we add some tests?

erikbosch:
I have now added some test cases, but they do not cover all possible VSS/Databroker Datatypes. We have quite much duplication in code (and tests) due to the strong type checks in Rust, so not that easy to reuse. I think the level of tests are quite reasonable now.

databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -215,94 +217,170 @@ impl From<&proto::Datapoint> for broker::DataValue {
...
+            DataType::Int8
+            | DataType::Int16
+            | DataType::Int32
+            | DataType::Int64
+            | DataType::Int8Array
+            | DataType::Int16Array
+            | DataType::Int32Array
+            | DataType::Int64Array => {
+                let min_value = match metadata.min {
+                    Some(DataValue::Int32(value)) => Some(i64::from(value)),
```
rafaeling, line 259: Is it not possible in VSS to have `Int8` data types with `Int8` `min` and `max`?

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -1039,13 +928,189 @@ mod tests {
...
+        }
+    }
+
+    #[tokio::test]
+    async fn test_update_entries_min_max_int8() {
+        let broker = DataBroker::default();
+
+        let timestamp = std::time::SystemTime::now();
+
+        match helper_add_int8(&broker, "test.datapoint1", -6, timestamp).await {
```
rafaeling, line 988: What about adding also `min` and `max` as parameters to this `helper_add_...` functions to make it easier to understand at first glance what should happen here? I often had to look the `min` and `max` values used above.

rafaeling (commented):
Looks good to me, apart from some general comments like why `DataValue` does not implement `Int8` and `Uint8`, are we upcasting values to `int32/64` and `uint32/64`?
Also looks good unit test coverage for kuksa.val.v2
Thanks Erik :)

databroker/src/grpc/kuksa_val_v2/conversions.rs
erikbosch: Indeed.

But In Databroker we never use smaller datatypes than 32 bits for integers, see https://github.com/eclipse-kuksa/kuksa-databroker/blob/81a15f1db4d0bdb25d69233bd9bba7e0693c4d51/databroker/src/types.rs#L59

The general Databroker approach (as I understand it) is to use the VSS Type as basis for checking min/max, but then store it internally in i32/i64/u32/u64.

databroker/src/types.rs
```diff
@@ -233,158 +233,21 @@ impl DataValue {
...
-                }
-            }
-            (DataValue::Int32(value), DataValue::Float(other_value)) => {
-                Ok(f64::from(*value) < f64::from(*other_value))
-            }
-            (DataValue::Int32(value), DataValue::Double(other_value)) => {
-                Ok(f64::from(*value) < *other_value)
-            }
+    pub fn greater_than_equal(&self, other: &DataValue) -> Result<bool, CastError> {
+        match self.greater_than(other) {
```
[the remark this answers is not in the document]
erikbosch: That is better a question for @argerus. I have some theories - one is that we do not "save" that much memory by using u32 instead of u16/u8 and i32 instead of i8/i16, another is that it simplifies implementation as we do not need to handle that many combinations.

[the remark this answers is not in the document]
lukasmittag: https://groups.google.com/g/protobuf-c/c/UzcyFvEOvMY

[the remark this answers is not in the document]
lukasmittag: https://protobuf.dev/programming-guides/proto3/#scalar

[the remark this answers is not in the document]
lukasmittag: there's no int8 in proto3

databroker/src/broker.rs
```diff
@@ -1301,6 +1457,8 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
         change_type: ChangeType,
         entry_type: EntryType,
         description: String,
+        min: Option<types::DataValue>,
```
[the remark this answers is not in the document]
erikbosch: It is tested indirectly by tests like `test_update_entries_min_max_int8` in `kuksa_val_v2`. One could argue that those tests better should be in this file as they do not rely on `kuksa_val_v2` features. The reason why they ended up there is that the first created used a helper method that already existed there.

Let me know if you think they should be moved.

databroker/src/grpc/kuksa_val_v2/val.rs
erikbosch: As of today all helper functions have "predefined" max/min-values, as I did not had any big need to let the limits vary. I could add min and max as args, but unless we introduce some more advanced argument builder it will just result in3 magic numbers unless your IDE can shed light on what each position means.

Please let me know if you want use to add them as arguments.


`helper_add_int8(&broker, "test.datapoint1", -10, 5, -6, timestamp).await `

databroker/src/types.rs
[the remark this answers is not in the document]
argerus: > That is better a question for @argerus. I have some theories - one is that we do not "save" that much memory by using u32 instead of u16/u8 and i32 instead of i8/i16, another is that it simplifies implementation as we do not need to handle that many combinations.

Exactly. It reduces the combinatorial explosion in the implementation and doesn't really cost anything memory wise since they are all part of the same `enum` anyway. Using 32 & 64 bit integers might even be marginally more efficient cpu wise (probably doesn't matter much). And as @lukasmittag mentions, it would need to be converted anyway when using protobuf.

The VSS data type is enforced when setting a value, and is not encoded by the value enum itself. This simplifies setting a value to `NotAvailable` as well.

databroker/src/grpc/kuksa_val_v2/conversions.rs
argerus: > The general Databroker approach (as I understand it) is to use the VSS Type as basis for checking min/max, but then store it internally in i32/i64/u32/u64.

Exactly

databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -215,94 +217,170 @@ impl From<&proto::Datapoint> for broker::DataValue {
...
             description: Some(metadata.description.clone()),
             comment: None,
             deprecation: None,
             unit: metadata.unit.clone(),
-            value_restriction: match metadata.allowed.as_ref() {
-                Some(allowed) => match allowed {
-                    broker::DataValue::StringArray(vec) => Some(proto::ValueRestriction {
+
+            // We add value_restriction below
+            value_restriction: None,
```
argerus, line 230: I think it might be a bit easier to follow if you moved the conversion to a `value_restriction` to a separate function and just assigned it here.

databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -215,94 +217,170 @@ impl From<&proto::Datapoint> for broker::DataValue {
...
+            | DataType::Int64
+            | DataType::Int8Array
+            | DataType::Int16Array
+            | DataType::Int32Array
+            | DataType::Int64Array => {
+                let min_value = match metadata.min {
+                    Some(DataValue::Int32(value)) => Some(i64::from(value)),
+                    Some(DataValue::Int64(value)) => Some(value),
+                    // Assumption here that we already have checked types
+                    // so wither Int64 or None
```
argerus, line 262: Can you update the wording of all of these comments a bit?

"so wither Int64 or None" isn't that understandable. Also copy / pasted below a few times.

argerus (changes requested):

databroker/src/grpc/kuksa_val_v2/conversions.rs
erikbosch: Yep, it is coming. Intention is to show that it never shall be anything else than Int32/Int64/None, so safe to assume that `_` never will be anything else than `None`.

databroker/src/types.rs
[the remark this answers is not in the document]
rafaeling: Got it, thanks

databroker/src/broker.rs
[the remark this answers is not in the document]
rafaeling: Well if the implementation test is covered somewhere else it is also fine for me.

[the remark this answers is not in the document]
argerus: > One could argue that those tests better should be in this file as they do not rely on kuksa_val_v2 features.

Yeah, I think it should be tested here as well, as this is where the functionality is implemented (independent of the abstractions above it).

The tests in kuksa_val_v2 should test the responses at that level of abstraction, i.e. returned error codes etc.

databroker/src/broker.rs
```diff
@@ -417,12 +438,53 @@ impl Entry {
         }
     }
 
+    /// Returns true if min/max conditions is violated
+    /// value is supposed to be of the base type, like Int32 for an Int32Array
+    fn validate_value_min_max_violated(&self, value: &DataValue) -> bool {
```
argerus, line 443: The name of this function in combination with a boolean return value makes it quite likely to be misunderstood in my opinion.

I think it would be safer to rename it and have it return a `Result<(), ValidationError>` or something.. That would make it hard to misunderstand.

databroker/src/broker.rs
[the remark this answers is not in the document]
erikbosch: Tests now moved, but added a new one testing grpc error code

databroker/src/broker.rs
erikbosch: Refactored

rafaeling (approved):

PR #81, opened 2024-10-15
Subscribe by id Integer32 implementation

rafaeling:
A variation of the Subscribe method which improves performance by using an Integer32 ID instead of a string for signals parameter.

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -200,6 +200,8 @@ pub struct NotificationError {}
 
 #[derive(Debug, Clone, Default)]
 pub struct EntryUpdate {
+    pub id: Option<i32>,
+
     pub path: Option<String>,
 
     pub datapoint: Option<Datapoint>,
@@ -757,6 +759,7 @@ impl ChangeSubscription {
                                             let mut update = EntryUpdate::default();
                                             let mut notify_fields = HashSet::new();
                                             // TODO: Perhaps make path optional
+                                            update.id = Some(entry.metadata.id);
                                             update.path = Some(entry.metadata.path.clone());
                                             if changed_fields.contains(&Field::Datapoint)
                                                 && fields.contains(&Field::Datapoint)
@@ -813,6 +816,7 @@ impl ChangeSubscription {
                                 let mut update = EntryUpdate::default();
                                 let mut notify_fields = HashSet::new();
                                 // TODO: Perhaps make path optional
+                                update.id = Some(entry.metadata.id);
                                 update.path = Some(entry.metadata.path.clone());
                                 if fields.contains(&Field::Datapoint) {
                                     update.datapoint = Some(entry.datapoint.clone());
@@ -2010,6 +2014,7 @@ mod tests {
             .update_entries([(
                 id1,
                 EntryUpdate {
+                    id: Some(id1),
                     path: None,
                     datapoint: Some(Datapoint {
                         ts: time1,
@@ -2042,6 +2047,7 @@ mod tests {
             .update_entries([(
                 id1,
                 EntryUpdate {
+                    id: Some(id1),
                     path: None,
                     datapoint: Some(Datapoint {
                         ts: time1,
@@ -2064,6 +2070,7 @@ mod tests {
             .update_entries([(
                 id2,
                 EntryUpdate {
+                    id: Some(id2),
                     path: None,
                     datapoint: None,
                     actuator_target: Some(Some(Datapoint {
@@ -2136,6 +2143,7 @@ mod tests {
             .update_entries([(
                 id1,
                 EntryUpdate {
+                    id: Some(id1),
                     path: None,
                     datapoint: Some(Datapoint {
                         ts: SystemTime::now(),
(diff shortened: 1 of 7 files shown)


databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -1830,6 +1909,51 @@ mod tests {
         }
     }
 
+    async fn publish_value(
```
lukasmittag, line 1912: why is this part of the PR? Isn't it there yet? Has it changed?

lukasmittag (commented):
Looks good not tested yet. Just one comment. Will try to test and start today as well.

databroker/src/broker.rs
```diff
@@ -2010,6 +2014,7 @@ mod tests {
             .update_entries([(
                 id1,
                 EntryUpdate {
+                    id: Some(id1),
```
erikbosch, line 2017: Is this needed as we actually give `id1` a few line above as well. Like before we managed without the path.

argerus (changes requested):
This is PR is changing way more than necessary. Databroker already supports the concept of using `id` instead `path` to refer to signals.

This is for example used by providers in the `sdv.databroker.v1` interface.

To add support for this in `kuksa.val.v2`, you should in theory only need to change stuff under `grpc/kuksa_val_v2`.

erikbosch (dismissed):
LGTM

databroker/src/broker.rs
argerus: > Is this needed as we actually give id1 a few line above as well. Like before we managed without the path.

Yes, these changes should not be necessary.

rafaeling: If `id` is not stored in `EntryUpdate` how is it going to return in the response `map` the corresponding `id` of the Datapoint?
It is not possible to use the request map ids order since the number of signals that might changed and will be returned to user might be dynamic right?. how could I get its ìd` then to build the response?

argerus: > If `id` is not stored in `EntryUpdate` how is it going to return in the response 

By putting this information in the `ChangeNotification` instead.

You are right that the current implementation of `ChangeSubscription::notify` is (ab)using `EntryUpdate` to pass the path to the subscribers, even though it shouldn't.

```rust
// TODO: Perhaps make path optional
update.path = Some(entry.metadata.path.clone());
```

Compare with how `fn update(&mut self, id: i32, update: EntryUpdate)` and `QuerySubscription::notify` is not using `EntryUpdate` to communicate this information.

databroker/src/grpc/kuksa_val_v2/val.rs
rafaeling: That was test specific, I placed it outside to be used by `test_subscribe_case` and `test_subscribe_case_by_id`

databroker/src/broker.rs
rafaeling: Updated

erikbosch (approved):
LGTM

erikbosch:
@argerus  - can you check if you are happy with it now

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -1907,27 +2116,37 @@ mod tests {
...
-            match broker.publish_value(request).await {
-                Ok(response) => {
-                    // Handle the successful response
-                    let publish_response = response.into_inner();
+                },
+            );
 
-                    // Check if there is an error in the response
-                    assert_eq!(publish_response, proto::PublishValueResponse {});
+            let f = false;
```
argerus, line 2122: What's with the `let f = false;`?

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -1907,27 +2116,37 @@ mod tests {
...
+                    // Process the SubscribeResponse
+                    let response = &subscribe_response.entries;
+                    assert_eq!(response.len(), expected_response.len());
+                    for key in response.keys() {
+                        match (response.get(key), expected_response.get(key)) {
+                            (Some(entry1), Some(entry2)) => {
+                                assert_eq!(entry1.value, entry2.value);
+                            }
+                            (Some(entry1), None) => {
+                                assert!(f, "Key '{}' is only in response: {:?}", key, entry1)
```
argerus, line 2134: Might as well use `panic!`

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -1883,21 +1998,115 @@ mod tests {
...
         }
 
-        async fn publish_value(
-            broker: &DataBroker,
-            entry_id: i32,
-            input_value: Option<bool>,
-            input_timestamp: Option<std::time::SystemTime>,
-        ) {
-            let timestamp = input_timestamp.map(|input_timestamp| input_timestamp.into());
+        let f = false;
```
argerus, line 2001: Why the `let f = false;`?

argerus (changes requested):

databroker/src/grpc/kuksa_val_v2/val.rs
rafaeling: We have been dragging this kind of assertion for a while I think.

argerus (approved):

PR #83, opened 2024-10-16
Implement server info service method

rafaeling:


databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -856,7 +856,12 @@ impl proto::val_server::Val for broker::DataBroker {
         &self,
         _request: tonic::Request<proto::GetServerInfoRequest>,
     ) -> Result<tonic::Response<proto::GetServerInfoResponse>, tonic::Status> {
-        Err(tonic::Status::unimplemented("Unimplemented"))
+        let server_info = proto::GetServerInfoResponse {
```
erikbosch, line 859: Should we possibly have a test case where we actually call `get_server_info` and check the results?

rafaeling: Added

erikbosch (approved):
LGTM

PR #85, opened 2024-10-17
Dep Update

erikbosch:
Possibly fixes build error

Changed by `cargo update`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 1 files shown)


erikbosch:
@argerus @rafaeling - I noticed that unit tests on your PRs failed as well as on master. Updating dependencies seems to solve it. Could it possibly be that `llvm` needed a newer `libc`?

lukasmittag:
Why only update Cargo.lock and not the toml as well?

lukasmittag (approved):
LGTM

PR #87, opened 2024-10-24
Update incomplete information on errors

rafaeling:


databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -505,34 +509,18 @@ impl proto::val_server::Val for broker::DataBroker {
...
+    //   NOT_FOUND if any of the signals are non-existant.
+    //   PERMISSION_DENIED
+    //       - if access is denied for any of the signals.
+    //   UNAUTHENTICATED if no credentials provided or credentials has expired
+    //   OUT_OF_RANGE is the provider is out of the min/max range specified
+    //   INVALID_ARGUMENT
+    //       - if the data type used in the request does not match
+    //            the data type of the addressed signal
+    //       - if the published value is not accepted,
+    //            e.g. if sending an unsupported enum value
```
erikbosch, line 522: Is this allowed value check? Or does that give out of range error?

proto/kuksa/val/v2/val.proto
```diff
@@ -45,30 +45,44 @@ service VAL {
...
   // Subscribe to a set of signals using i32 id parameters
   // Returns (GRPC error code):
   //   NOT_FOUND if any of the signals are non-existant.
+  //   UNAUTHENTICATED if no credentials provided or credentials has expired
   //   PERMISSION_DENIED if access is denied for any of the signals.
+  //   INVALID_ARGUMENT if the request is empty
+  //
+  // When subscribing the Broker shall immediately return the value for all
+  // subscribed entries. If no value is available when subscribing a DataPoint
+  // with value None shall be returned.
```
erikbosch, line 67: Maybe rephrase last sentence

proto/kuksa/val/v2/val.proto
```diff
@@ -112,7 +135,32 @@ service VAL {
...
-  // Errors are communicated as messages in the stream.
+  //
+  // Errors:
+  //    - Provider sends ProvideActuationRequest -> Databroker returns ProvideActuationResponse
+  //        Returns (GRPC error code) and closes the stream call (strict case).
+  //          NOT_FOUND if any of the signals are non-existant.
+  //          PERMISSION_DENIED if access is denied for any of the signals.
+  //          UNAUTHENTICATED if no credentials provided or credentials has expired
+  //          ALREADY_EXISTS if a provider already claimed the ownership of an actuator
+  //          UNAUTHENTICATED if permission expired
```
erikbosch, line 146: Duplicate

databroker/src/grpc/kuksa_val_v2/val.rs
rafaeling: It seems so, https://github.com/boschglobal/kuksa-databroker/blob/fix/doc_open_provider_stream/databroker/src/broker.rs#L434

The issue I see now is that we map to some different error but not to `INVALID_ARGUMENT` https://github.com/boschglobal/kuksa-databroker/blob/fix/doc_open_provider_stream/databroker/src/grpc/kuksa_val_v2/conversions.rs#L462

rafaeling: Also fixed now

erikbosch: Looking at our code we should maybe fix some conversions of OutOfBounds.

```
broker::UpdateError::OutOfBounds => tonic::Status::new(
                tonic::Code::OutOfRange,
                format!("Value out of bounds (id: {})", id),
            ),
```

According to https://grpc.github.io/grpc/cpp/md_doc_statuscodes.html `OutOfRange` seems rather to be intended for state-related errors like trying to read the forth item in an array but there are only 3, not for cases where we fail a static check like checking that argument has correct type and fulfills min/max/allowed.

So maybe refactor, remove OUT_OF_RANGE for at least v2 and describe all errors under "invalid argument", i.e. wrong type, min/max exceeded, allowed not fulfilled.

rafaeling: Done

erikbosch (approved):
LGTM

PR #88, opened 2024-10-28
Fix actuate and batch actuate out of range error

rafaeling:


databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -2476,6 +2380,61 @@ mod tests {
...
+            }),
+        });
+
+        request
+            .extensions_mut()
+            .insert(permissions::ALLOW_ALL.clone());
+
+        let result_response = proto::val_server::Val::actuate(&broker, request).await;
+        assert!(result_response.is_err());
+        assert_eq!(result_response.unwrap_err().code(), tonic::Code::OutOfRange)
```
wba2hi, line 2435: grafik
I think the last two commits here are incompatible / create merge conflicts. Either needs to be changed after the other is merged. tonic::Code::OutOfRange needs to be changed to InvalidValue

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -2598,6 +2557,126 @@ mod tests {
...
+            ],
+        });
+
+        request
+            .extensions_mut()
+            .insert(permissions::ALLOW_ALL.clone());
+
+        let result_response = proto::val_server::Val::batch_actuate(&broker, request).await;
+        assert!(result_response.is_err());
+        assert_eq!(result_response.unwrap_err().code(), tonic::Code::OutOfRange)
```
wba2hi, line 2677: see previous finding

wba2hi (commented):
Added some comments

databroker/src/grpc/kuksa_val_v2/val.rs
rafaeling: Yes you are right, I branched from the branch `fix/doc_open_provider_stream` I need to change it to make it work

databroker/src/broker.rs
```diff
@@ -1882,6 +1885,58 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+                    Err(UpdateError::UnsupportedType) => {
+                        let message = format!("Unsupported type for vss_id {}", vss_id);
+                        Err((ActuationError::UnsupportedType, message))
+                    }
+                    Err(UpdateError::WrongType) => {
+                        let message = format!("Wrong type for vss_id {}", vss_id);
+                        Err((ActuationError::WrongType, message))
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
rafaeling: I have done my best here but some information is not included since it is missing at this level.

databroker/src/broker.rs
```diff
@@ -1882,6 +1885,71 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+        match result_entry {
+            Ok(entry) => {
+                let validation = entry.validate_value(data_value);
+                let vss_path = entry.metadata.path;
+                match validation {
+                    Ok(_) => Ok(()),
+                    Err(UpdateError::NotFound) => {
+                        let message = format!("Could not resolve vss_path {}", vss_path);
+                        Err((ActuationError::NotFound, message))
+                    }
```
wba2hi, line 1904: Might be redundant. 
NotFound is thrown by get_entry_by_id which we use in the method itself but also here in line 1894 for the vss_id lookup. Since we are already in the OK case even a consecutive lookup of the id should not lead to NotFound.

databroker/src/broker.rs
```diff
@@ -1882,6 +1885,71 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+                    Err(UpdateError::OutOfBounds) => {
+                        let message = format!(
+                            "Out of bounds value provided for {}: {} | Expected range [min: {}, max: {}]",
+                            vss_path,
+                            data_value,
+                            entry.metadata.min.unwrap(),
+                            entry.metadata.max.unwrap()
```
wba2hi, line 1911: This might lead to a crash when min or max is not specified, unfortunately some percentual paths in the vss spec only have one. e.g. Vehicle.Chassis.Axle.Row1.Wheel.Left.Brake.PadWear

wba2hi: Don't put to much effort into this - as the logic might change when being returned as an "InvalidArgument" anyway.

databroker/src/broker.rs
rafaeling: `PermissionDenied` and `PermissionExpired` also are redundant, I left to make Rust happy but it just maybe enough if we leave `UnsupportedType`, `WrongType` and `OutOfBounds`

databroker/src/broker.rs
rafaeling: Done

databroker/src/broker.rs
wba2hi: Unfortunately I'm not totally satisfied with the current solution :( Now as a default case we cause a panic, which is - in our controlled scenario - totally fine. However, due to the use of the default branch we will not see when a new UpdateError case is added and therefore create a scenario in which a panic is thrown for the newly added UpdateError during conversion to the ActuationError because it was forgotten to be added here. 

Please find a better solution for the default case - in which we don't panic - or change it back to an exhaustive match (with tolerated redundancy - maybe add a comment to show, that the redundancy is accepted). Using an exhaustive match, will create compiler errors in case of an UpdateError being added and not adding a new case to the match - therefore the developer easily sees, that he needs to add it here. 

Sorry for the back and forth!

rafaeling: I will leave the extended version with a comment, maybe the `validate_actuator_value` method could be use somewhere else or even at some point `get_entry_by_id` also modified, who knows, it does not hearth to have errors I think.

wba2hi (approved):
LGTM

PR #89, opened 2024-10-29
Add reflection support for kuksa.val.v2

rafaeling:
Release binaries generated locally went from 6.3 MB to 6.5 MB after reflection support was added for kuksa.val.v2

It helps to test kuksa.val.v2 with grpcurl until databroker-cli is updated, i.e:

grpcurl -d '{  "signal_id": { "path": "Vehicle.Speed" },  "data_point": { "timestamp": "2024-10-29T12:04:31.463835877Z", "value": { "float": 75.5 }  }}' -plaintext localhost:55555 kuksa.val.v2.VAL.PublishValue
 
grpcurl -d '{ "signal_id": { "path": "Vehicle.Speed" } }' -plaintext localhost:55555 kuksa.val.v2.VAL.GetValue
 
grpcurl -d '{ "signal_paths": ["Vehicle.Speed"]}' -plaintext localhost:55555 kuksa.val.v2.VAL.Subscribe

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,7 @@ dependencies = [
  "tokio-stream",
  "tonic",
  "tonic-mock",
+ "tonic-reflection",
  "tracing",
  "tracing-subscriber",
  "uuid",
@@ -3138,6 +3139,19 @@ dependencies = [
  "http",
  "http-body",
  "prost",
+ "tonic",
+]
+
+[[package]]
+name = "tonic-reflection"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "548c227bd5c0fae5925812c4ec6c66ffcfced23ea370cb823f4d18f0fc1cb6a7"
+dependencies = [
+ "prost",
+ "prost-types",
+ "tokio",
+ "tokio-stream",
  "tonic",
 ]
 
--- a/databroker-proto/build.rs
+++ b/databroker-proto/build.rs
@@ -11,6 +11,8 @@
 * SPDX-License-Identifier: Apache-2.0
 ********************************************************************************/
 
+use std::{env, path::PathBuf};
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     std::env::set_var("PROTOC", protobuf_src::protoc());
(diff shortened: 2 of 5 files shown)


lukasmittag:
If we enable it shouldn't we enable it for the other interfaces as well not only v2?

erikbosch:
Should we possibly add some info on how this can be used, either in the wiki (https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Testing ? ) or somewhere in https://github.com/eclipse-kuksa/kuksa-databroker/tree/main/doc

rafaeling:
> Should we possibly add some info on how this can be used, either in the wiki (https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Testing ? ) or somewhere in https://github.com/eclipse-kuksa/kuksa-databroker/tree/main/doc

I think it would be good to add it to the Release-Testing to ensure all errors are returned correctly.

erikbosch (approved):
Looks good to me. For me it is OK that we only have discovery for the new API. I tested with the command below and it reported what I expected.

```
erik@debian6:~/kuksa.val$ docker run --network=host fullstorydev/grpcurl  --plaintext -vv 127.0.0.1:55555 describe
kuksa.val.v2.VAL is a service:
service VAL {
```

But my approval is conditioned on that we add a short section in the release testing wiki, possibly just stating that you should run a command like above. Even better would be if we could have an automatic test, but that would require more work I think

rafaeling:
> Looks good to me. For me it is OK that we only have discovery for the new API. I tested with the command below and it reported what I expected.
> 
> ```
> erik@debian6:~/kuksa.val$ docker run --network=host fullstorydev/grpcurl  --plaintext -vv 127.0.0.1:55555 describe
> kuksa.val.v2.VAL is a service:
> service VAL {
> ```
> 
> But my approval is conditioned on that we add a short section in the release testing wiki, possibly just stating that you should run a command like above. Even better would be if we could have an automatic test, but that would require more work I think

@erikbosch there it is https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Testing#testing-databroker-kuksavalv2-api-methods-with-grpcurl

PR #90, opened 2024-10-30
Validation of actuator type on actuate and batch_actuate and some more errors

rafaeling:


databroker/src/broker.rs
```diff
@@ -1886,39 +1892,60 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
                     Err(UpdateError::OutOfBounds) => {
-                        let message = format!(
-                            "Out of bounds value provided for {}: {} | Expected range [min: {}, max: {}]",
-                            vss_path,
-                            data_value,
-                            entry.metadata.min.map_or("None".to_string(), |value| value.to_string()),
-                            entry.metadata.max.map_or("None".to_string(), |value| value.to_string()),
-                        );
-                        Err((ActuationError::OutOfBounds, message))
+                        let mut message = "No message defined".to_string();
```
erikbosch, line 1913: I am not really sure if we should try to be smart here and come up with the reason why we got "OutOfBounds". We do not only report OutOfBounds when min/max or allowed is exceeded,also in some other cases like when the VSS datatype is uint8 but Broker got a 275 as value.

If needed it would be better I think to split UpdateError::OutOfBounds to multiple ones indicating reason, like UpdateError::OutOfBoundsAllowed, UpdateError::OutOfBoundsMinMax, UpdateError::OutOfBoundsType

Or alternatively create a message deepr down in the stack

databroker/src/broker.rs
```diff
@@ -1886,39 +1892,60 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+                        if metadata.min.is_some() || metadata.max.is_some() {
+                            message = format!(
+                                "Out of bounds value provided for {}: {} | Expected range [min: {}, max: {}]",
+                                vss_path,
+                                data_value,
+                                metadata.min.map_or("None".to_string(), |value| value.to_string()),
+                                metadata.max.map_or("None".to_string(), |value| value.to_string()),
+                            );
+                        } else if metadata
```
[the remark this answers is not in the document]
wba2hi: Yes. allowed is an optional, therefore calling unwrap on it might lead to a panic if the option is set to None.

databroker/src/broker.rs
rafaeling: I prefer the approach of extending UpdateError and propagating the correct type of error upstream.
Thanks.

databroker/src/types.rs
```diff
@@ -485,6 +485,36 @@ impl DataValue {
             _ => Err(CastError {}),
         }
     }
+
+    pub fn contains(&self, other: &DataValue) -> Result<bool, CastError> {
```
erikbosch, line 489: This function is not used, right?

erikbosch (approved):
I think it is good enough now

PR #91, opened 2024-10-30
Build custom runtime for configurable worker threads

lukasmittag:
to make the worker threads configurable via env varibale and command line. Documentation currently missing!!!!

### DIFF
--- a/databroker/src/main.rs
+++ b/databroker/src/main.rs
@@ -345,7 +345,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         databroker::init_logging();
 
         info!("Starting Kuksa Databroker {}", version);
-        info!("Using {} threads with {} cores available on the system", worker_threads, cores);
+        info!(
+            "Using {} threads with {} cores available on the system",
+            worker_threads, cores
+        );
 
         let ip_addr = args.get_one::<String>("address").unwrap().parse()?;
         let port = args
--- a/databroker/src/main.rs
+++ b/databroker/src/main.rs
@@ -26,6 +26,7 @@ use tokio::signal::unix::{signal, SignalKind};
 #[cfg(feature = "tls")]
 use tracing::warn;
 use tracing::{debug, error, info};
+use std::thread::available_parallelism;
 
 use clap::{Arg, ArgAction, Command};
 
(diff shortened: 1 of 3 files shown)


databroker/src/main.rs
```diff
@@ -321,152 +331,162 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    )
-    .await;
-
-    if let Some(metadata_filenames) = args.get_many::<String>("vss-file") {
-        for filename in metadata_filenames {
-            read_metadata_file(&database, filename).await?;
-        }
-    }
+    let cores = available_parallelism().unwrap().get();
+    let worker_threads: &usize = args.get_one::<usize>("worker-threads").unwrap_or(&cores);
```
rafaeling, line 335: What if the number of threads provided is higher than the number of threads allowed?

lukasmittag: Whats meant by number of threads allowed? You can spawn more threads but if you do not have more cores than it will block the thread. We could limit it to the cores available but I do not think it is necessary.

rafaeling: I mean like for example, the user has no idea how many cores the test device has and provides a `woker-threads` parameter higher than the available number of cores.

erikbosch: As I understand it is nothing that prevents you from having more threads than cores, but you do not gain anything from it, but most likely you will get a small performance penalty if specifying more than number of cores. I think it makes sense to write some advice in documentation on when it makes sense to have something else than default.

Is a possible use-case that you want to run Databroker with high priority (for minimal latency), but at the same time you do not want it to steal more than a certain share of the capacity and by that reason you state that it only might use X cores.

erikbosch: Would there be a value to print something like "Using X cores, Y cores available", either on debug or info level? Info-level if we consider it as relevant info to get when someone has performance problems and we do not want to get a huge debug log.

lukasmittag: Added info message how many threads are spawned and how many cores are available on the system.

erikbosch (approved):
LGTM

Did some smoke test using `kuksa-client`. No problems detected


```
erik@debian6:~/kuksa-databroker$ cargo run --bin databroker -- --metadata data/vss-core/vss_release_4.0.json --insecure --worker-threads=2000
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/databroker --metadata data/vss-core/vss_release_4.0.json --insecure --worker-threads=2000`
2024-11-05T13:39:01.519395Z  INFO databroker: Init logging from RUST_LOG (environment variable not found)
2024-11-05T13:39:01.519456Z  INFO databroker: Starting Kuksa Databroker 0.4.7-dev.0
2024-11-05T13:39:01.519470Z  INFO databroker: Using 2000 threads with 3 cores available on the system

```

rafaeling (approved):
Looks good to me

PR #92, opened 2024-10-31
Adding opentelemety as a feature "otel"

LikhithST:
With this feature "otel" enabled, it conditionally compiles the opentelemetry and the related tracing functionalities.

### DIFF
--- a/databroker/src/open_telemetry.rs
+++ b/databroker/src/open_telemetry.rs
@@ -1,0 +1,21 @@
+use opentelemetry::{KeyValue, runtime};
+use opentelemetry::sdk::{Resource, trace};
+use opentelemetry::trace::TraceError;
+use opentelemetry_otlp::WithExportConfig;
+
+pub fn init_trace() -> Result<trace::Tracer, TraceError> {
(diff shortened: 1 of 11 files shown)


databroker/src/open_telemetry.rs
```diff
@@ -0,0 +1,25 @@
...
+};
+
+#[cfg(feature="otel")]
+pub fn init_trace() -> Result<trace::Tracer, TraceError> {
+    opentelemetry_otlp::new_pipeline()
+        .tracing()
+        .with_exporter(
+            opentelemetry_otlp::new_exporter()
+                .tonic()
+                .with_endpoint("http://localhost:4317"),
```
erikbosch, line 16: Is 4317 a standard port for this type of functionality?

LikhithST: Yes, Its a standard port for Jaeger backend. As Jaeger backend can also run on other device, I have made the endpoint configurable through env variable. 

This latest change will be part of my new signed PR (As this current PR is not signed, I was requested by Lukas to create a new signed PR)

LikhithST:
New PR will be created with signed commit

PR #93, opened 2024-11-04
Add missing min max restriction for sdv.databroker.v1

rafaeling:


proto/sdv/databroker/v1/types.proto
```diff
@@ -163,7 +151,49 @@ message Metadata {
   ChangeType change_type = 6;  // CONTINUOUS or STATIC or ON_CHANGE
   string description     = 7;
 
-  Allowed allowed = 10;
+  ValueRestriction value_restriction = 10;
```
erikbosch, line 154: Do we really want to introduce it to this old API?

rafaeling: If we don't add `min` and `max`, there is no way for users to know what is the correct range of values to publish, i.e databroker-perf (provider)

erikbosch: Yes, but they have managed with that so far

databroker/src/grpc/sdv_databroker_v1/conversions.rs
```diff
@@ -311,54 +457,7 @@ impl From<&broker::Metadata> for proto::Metadata {
...
-                | Some(broker::DataValue::String(_))
-                | Some(broker::DataValue::Int32(_))
-                | Some(broker::DataValue::Int64(_))
-                | Some(broker::DataValue::Uint32(_))
-                | Some(broker::DataValue::Uint64(_))
-                | Some(broker::DataValue::Float(_))
-                | Some(broker::DataValue::Double(_))
-                | None => None,
-            },
+            value_restriction: value_restriction_from(metadata),
```
erikbosch, line 460: If we add this we should better have one or more test cases as sanity check

proto/sdv/databroker/v1/types.proto
rafaeling: > Yes, but they have managed with that so far

I think before it worked since min, max were never checked for release 0.4.6, maybe even users never use min and max in the vss json file on databroker, however if we dont add this, we will not be able to generate correct values to measure the performance in case there are users using min max concept in their vss files.

erikbosch (approved):
LGTM

PR #94, opened 2024-11-05
Add support for unix domain sockets

argerus:


erikbosch:
I think we should add some documentation. Some things I think could be useful to include/mention:

- Does the CLI support it?
- Provide a example where we use the new argument. Preferably also with a client (if supported)
- Mention that the socket is not protected by TLS or similar, that you may need to consider to mechanisms like MAC/DAC for protecting the socket
- Mention that token functionality works the same way as for TCP/IP connection.

When merged we should preferably also mention something on how to test it in https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Testing - I think we at least shall have some form of smoke test, manual or in Github CI to verify that it works before releasing.

databroker/src/main.rs
```diff
@@ -457,7 +480,45 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         apis.push(grpc::server::Api::SdvDatabrokerV1);
     }
 
-    grpc::server::serve(
+    let unix_socket = args.get_one::<String>("unix-socket").cloned();
+    if let Some(path) = unix_socket {
+        unlink_unix_domain_socket(&path)?;
```
lukasmittag, line 485: why do we need to unlink first?

erikbosch: I think it is a quite common approach. Normally on a clean shutdown we will do the unlinking at exit, but for crashes and forced shutdown (`kill -9`) we will not unlink it, and then best to try to unlink it before creating it.

lukasmittag: understood

databroker/src/grpc/server.rs
```diff
@@ -138,23 +131,55 @@ where
     .await
 }
 
-pub async fn serve_with_incoming_shutdown<F>(
-    listener: TcpListener,
+pub async fn serve_uds<F>(
```
erikbosch, line 134: Do we want to get rid of `uds` everywhere, i.e. rename this one to `serve_unix_domain_socket` or similar?

databroker/src/main.rs
argerus: Also, it will fail to bind the path if there is already a file in place.

But as you rightfully point out, a comment explaining the reason _why_ this is done would be appropriate :smile:

erikbosch:
Rebased the PR and added some basic documentation. Also addressed two of the comments

erikbosch (approved):
LGTM

Issue #60, opened 2024-09-04
Slow subscriber might block publishes


jjj-vtm:
Hi, 

I wrote a client which basically simulates an overloaded (not consuming) subscriber. 

```
async fn main() -> () {
    let paths = vec!["Vehicle.Speed"];
    let mut client: KuksaClient =
        KuksaClient::new(kuksa_common::to_uri("127.0.0.1:55556").unwrap());

    let mut stream = client.subscribe(paths).await.unwrap();

    match stream.message().await {
        Ok(msg) => {
            println!("Got {:?}, not consuming any more message.", msg);
        }
        Err(_) => todo!(),
    }
    tokio::time::sleep(Duration::MAX).await; 
}
```

to trigger the bug faster I also modified the http/2 initial_connect_window and stream_window_size
```
async fn try_create_channel(&mut self) -> Result<&Channel, ClientError> {
        #[cfg(feature = "tls")]
        let mut builder = tonic::transport::Channel::builder(self.uri.clone());
        builder = builder.initial_connection_window_size(2048);
        builder = builder.initial_stream_window_size(2048);
```

After around 15-20 publishes via the cli 

kuksa.val.v1 > publish Vehicle.Speed 123
[publish]  OK

no response will come back and publishing a different value via another cli session 
kuksa.val.v1 > publish Vehicle.Width 3

will also hang. I guess that this send 

https://github.com/eclipse-kuksa/kuksa-databroker/blob/ccc382c042be28e015cdfad170e56deb6bacdca6/databroker/src/broker.rs#L752

will wait until the receiver reads the data and since a read lock on database is held subsequent publishes wait on the lock. But even with a different database implementation the issue would be that one slow consumer might cause delays in the value propagation for other subscribers on the same path since the send from above is called in a loop over the possible subscribers. 

Maybe it would be better to use try_send instead of send. 

Cheers, 

Jan

argerus:
Nice catch!

I've actually had this issue in the back of my mind for some time now, but haven't gotten around to fixing / documenting it.

> Maybe it would be better to use try_send instead of send.

[...]

PR #95, opened 2024-11-05
Use tokio broadcast::channel

rafaeling:
Issue -> https://github.com/eclipse-kuksa/kuksa-databroker/issues/60

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -18,8 +18,9 @@ use crate::query;
 pub use crate::types::{ChangeType, DataType, DataValue, EntryType};
 
 use tokio::sync::{broadcast, mpsc, RwLock};
+use tokio_stream::wrappers::BroadcastStream;
 use tokio_stream::wrappers::ReceiverStream;
-use tokio_stream::Stream;
+use tokio_stream::{Stream, StreamExt};
 
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
@@ -32,6 +33,8 @@ use crate::types::ExecutionInputImplData;
 use tracing::{debug, info, warn};
 
 use crate::glob;
+
+const MAX_SUBSCRIBE_BUFFER_SIZE: usize = 1000;
 
 #[derive(Debug)]
 pub enum ActuationError {
@@ -135,14 +138,14 @@ pub struct QueryField {
     pub value: DataValue,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ChangeNotification {
     pub id: i32,
     pub update: EntryUpdate,
     pub fields: HashSet<Field>,
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
 pub struct EntryUpdates {
     pub updates: Vec<ChangeNotification>,
 }
@@ -157,6 +160,7 @@ pub enum QueryError {
 pub enum SubscriptionError {
     NotFound,
     InvalidInput,
+    InvalidBufferSize,
     InternalError,
 }
 
@@ -198,7 +202,7 @@ pub struct QuerySubscription {
 
 pub struct ChangeSubscription {
     entries: HashMap<i32, HashSet<Field>>,
-    sender: mpsc::Sender<EntryUpdates>,
+    sender: broadcast::Sender<EntryUpdates>,
     permissions: Permissions,
 }
 
@@ -805,7 +809,7 @@ impl Subscriptions {
             }
         });
         self.change_subscriptions.retain(|sub| {
-            if sub.sender.is_closed() {
+            if sub.sender.receiver_count() == 0 {
                 info!("Subscriber gone: removing subscription");
                 false
             } else if sub.permissions.is_expired() {
@@ -898,9 +902,12 @@ impl ChangeSubscription {
                     if notifications.updates.is_empty() {
                         Ok(())
                     } else {
-                        match self.sender.send(notifications).await {
-                            Ok(()) => Ok(()),
-                            Err(_) => Err(NotificationError {}),
+                        match self.sender.send(notifications) {
+                            Ok(_number_of_receivers) => Ok(()),
+                            Err(err) => {
+                                debug!("Send error for entry{}: ", err);
+                                Err(NotificationError {})
+                            }
                         }
                     }
                 } else {
@@ -939,9 +946,12 @@ impl ChangeSubscription {
                     }
                     notifications
                 };
-                match self.sender.send(notifications).await {
-                    Ok(()) => Ok(()),
-                    Err(_) => Err(NotificationError {}),
+                match self.sender.send(notifications) {
+                    Ok(_number_of_receivers) => Ok(()),
+                    Err(err) => {
+                        debug!("Send error for entry{}: ", err);
+                        Err(NotificationError {})
+                    }
                 }
             }
         }
@@ -1622,12 +1632,22 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
     pub async fn subscribe(
         &self,
         valid_entries: HashMap<i32, HashSet<Field>>,
+        buffer_size: Option<usize>,
     ) -> Result<impl Stream<Item = EntryUpdates>, SubscriptionError> {
         if valid_entries.is_empty() {
             return Err(SubscriptionError::InvalidInput);
         }
 
-        let (sender, receiver) = mpsc::channel(10);
+        let channel_capacity = if let Some(cap) = buffer_size {
+            if cap > MAX_SUBSCRIBE_BUFFER_SIZE {
+                return Err(SubscriptionError::InvalidBufferSize);
+            }
+            cap
+        } else {
+            1
+        };
+
+        let (sender, receiver) = broadcast::channel(channel_capacity);
         let subscription = ChangeSubscription {
             entries: valid_entries,
             sender,
@@ -1648,7 +1668,13 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
             .await
             .add_change_subscription(subscription);
 
-        let stream = ReceiverStream::new(receiver);
+        let stream = BroadcastStream::new(receiver).filter_map(|result| match result {
+            Ok(message) => Some(message),
+            Err(err) => {
+                debug!("Lagged entries: {}", err);
+                None
+            }
+        });
         Ok(stream)
     }
 
@@ -4234,20 +4260,23 @@ pub mod tests {
             .expect("Register datapoint should succeed");
 
         let mut stream = broker
-            .subscribe(HashMap::from([(id1, HashSet::from([Field::Datapoint]))]))
+            .subscribe(
+                HashMap::from([(id1, HashSet::from([Field::Datapoint]))]),
+                None,
+            )
             .await
             .expect("subscription should succeed");
 
         // Stream should yield initial notification with current values i.e. NotAvailable
         match stream.next().await {
-            Some(next) => {
-                assert_eq!(next.updates.len(), 1);
+            Some(entry) => {
+                assert_eq!(entry.updates.len(), 1);
                 assert_eq!(
-                    next.updates[0].update.path,
+                    entry.updates[0].update.path,
                     Some("test.datapoint1".to_string())
                 );
                 assert_eq!(
-                    next.updates[0].update.datapoint.as_ref().unwrap().value,
+                    entry.updates[0].update.datapoint.as_ref().unwrap().value,
                     DataValue::NotAvailable
                 );
             }
@@ -4281,14 +4310,14 @@ pub mod tests {
 
         // Value has been set, expect the next item in stream to match.
         match stream.next().await {
-            Some(next) => {
-                assert_eq!(next.updates.len(), 1);
+            Some(entry) => {
+                assert_eq!(entry.updates.len(), 1);
                 assert_eq!(
-                    next.updates[0].update.path,
+                    entry.updates[0].update.path,
                     Some("test.datapoint1".to_string())
                 );
                 assert_eq!(
-                    next.updates[0].update.datapoint.as_ref().unwrap().value,
+                    entry.updates[0].update.datapoint.as_ref().unwrap().value,
                     DataValue::Int32(101)
                 );
             }
--- a/proto/kuksa/val/v2/val.proto
+++ b/proto/kuksa/val/v2/val.proto
@@ -52,12 +52,18 @@ service VAL {
   //   NOT_FOUND if any of the signals are non-existant.
   //   UNAUTHENTICATED if no credentials provided or credentials has expired
   //   PERMISSION_DENIED if access is denied for any of the signals.
-  //   INVALID_ARGUMENT if the request is empty or provided path is too long
-  //       - MAX_REQUEST_PATH_LENGTH: usize = 1000;
+  //   INVALID_ARGUMENT
+  //       - if the request is empty or provided path is too long
+  //             MAX_REQUEST_PATH_LENGTH: usize = 1000;
+  //       - if buffer_size exceeds the maximum permitted
+  //             MAX_BUFFER_SIZE: usize = 1000;
   //
   // When subscribing, Databroker shall immediately return the value for all
   // subscribed entries.
   // If a value isn't available when subscribing to a it, it should return None
+  //
+  // If a subscriber is slow to consume signals, messages will be buffered up
+  // to the specified buffer_size before the oldest messages are dropped.
   //
   rpc Subscribe(SubscribeRequest) returns (stream SubscribeResponse);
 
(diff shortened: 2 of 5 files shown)


databroker/src/broker.rs
```diff
@@ -1627,7 +1628,7 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
             return Err(SubscriptionError::InvalidInput);
         }
 
-        let (sender, receiver) = mpsc::channel(10);
+        let (sender, receiver) = broadcast::channel(100);
```
lukasmittag, line 1631: "You're right that it's probably "overkill" to use one broadcast channel per subscriber, but the other nice semantic of the broadcast channel is that it will drop the oldest value when the subscriber is lagging." (John Argerus)

Have you put a thought into this?

databroker/src/broker.rs
```diff
@@ -2054,7 +2062,7 @@ pub mod tests {
     use crate::permissions;
 
     use super::*;
-    use tokio_stream::StreamExt;
+    use futures::StreamExt;
```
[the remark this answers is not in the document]
rafaeling: Fixed

databroker/src/broker.rs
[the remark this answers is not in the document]
rafaeling: Outdated

databroker/src/broker.rs
```diff
@@ -898,8 +899,8 @@ impl ChangeSubscription {
                     if notifications.updates.is_empty() {
                         Ok(())
                     } else {
-                        match self.sender.send(notifications).await {
-                            Ok(()) => Ok(()),
+                        match self.sender.send(notifications) {
+                            Ok(_) => Ok(()),
```
argerus, line 903: Unless there's a good reason, it's better to match explicitly (so a potential change in the return values of `send`, however unlikely, would be caught by the compiler).

databroker/src/broker.rs
```diff
@@ -939,8 +940,8 @@ impl ChangeSubscription {
                     }
                     notifications
                 };
-                match self.sender.send(notifications).await {
-                    Ok(()) => Ok(()),
+                match self.sender.send(notifications) {
+                    Ok(_) => Ok(()),
```
argerus, line 944: Explicit is better than a catch all.

databroker/src/broker.rs
```diff
@@ -1648,7 +1649,13 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
             .add_change_subscription(subscription);
 
-        let stream = ReceiverStream::new(receiver);
+        let stream = BroadcastStream::new(receiver).filter_map(|result| match result {
+            Ok(message) => Some(message),
+            Err(err) => {
+                debug!("Lagged entries: {}", err);
+                None
+            }
+        });
```
argerus, line 1658: This should be forwarded to to the subscriber (of the `broadcast` channel), in order for them to be able to act on it.

databroker/src/broker.rs
```diff
@@ -4279,23 +4281,23 @@ pub mod tests {
...
-                );
-            }
-            None => {
-                panic!("did not expect stream end")
-            }
-        }
+        let stream = broker
+            .subscribe(HashMap::from([(id1, HashSet::from([Field::Datapoint]))]))
+            .await
+            .expect("subscription should succeed");
```
argerus, line 4287: Is this adding a re-subscription? Why?

databroker/src/broker.rs
```diff
@@ -4233,28 +4240,23 @@ pub mod tests {
...
+            assert_eq!(next.updates.len(), 1);
+            assert_eq!(
+                next.updates[0].update.path,
+                Some("test.datapoint1".to_string())
+            );
+            assert_eq!(
+                next.updates[0].update.datapoint.as_ref().unwrap().value,
+                DataValue::NotAvailable
+            );
+        });
```
argerus, line 4259: What's the reasoning behind the change here?

databroker/src/broker.rs
```diff
@@ -1627,7 +1628,7 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
             return Err(SubscriptionError::InvalidInput);
         }
 
-        let (sender, receiver) = mpsc::channel(10);
+        let (sender, receiver) = broadcast::channel(10);
```
argerus, line 1631: Since we're moving to a model of dropping notifications when a subscriber lags, I think the buffer size should be configurable by the subscriber (i.e. the caller of `pub async fn subscribe(...)`)

In general, I think it's most likely that subscribers are only interested in receiving the latest value, and don't care whether an intermediate value was dropped (i.e. `broadcast::channel(1)`). But since there _can_ be other use cases, it should probably be configurable.

For the previous model, i.e. nothing was to be dropped, the buffer acted as a way to not immediately put back pressure on the provider.

argerus (changes requested):

databroker/src/broker.rs
argerus: Since we're moving to a model of dropping notifications when a subscriber lags, I think the buffer size should be configurable by the subscriber (i.e. the caller of `pub async fn subscribe(...)`), instead of hard coding it.

I think it's most likely that subscribers are only interested in receiving the latest value without caring whether an intermediate value was dropped (i.e. `broadcast::channel(1)`). But since there _can_ be other use cases it should probably be configurable.

For the previous model, i.e. nothing was to be dropped, the buffer acted as a way to not immediately put back pressure on the provider.

rafaeling: I suppose you mean only to add the size field parameter to `kuksa.val.v2` subscribe request, but what would be the appropriate default value to be used then for `kuksa.val.v1` and `sdv.databroker.v1`?

erikbosch:
Do not know how much it says, but I did a smoke test with a subscription with kuksa-client and it worked without problem.

databroker/src/broker.rs
rafaeling: The size value returned by `send` method represents the number of currently subscribed Receiver handles at the time of the send, my question is, why is that relevant? how should we handle? just output in `debug`?

databroker/src/broker.rs
rafaeling: Don't know if I fully get that, you mean transmit the lagged error as tonic error to the consumer application?

databroker/src/broker.rs
erikbosch: What is the value range we expect. Like is 0 an expected response, or should it always be 1 or more?

rafaeling: After doing a test it seems to be always 1, since each new subscription creates its own broadcast::channel and senders are independent from each other.

argerus: I guess the return value actually doesn't matter in this case. But when putting it behind a `_` it's not evident what is returned.

> Unless there's a good reason, it's better to match explicitly [...]

I'm ok with ignoring it.

databroker/src/broker.rs
argerus: I guess the default at the moment ought to be `1` as that would be the better choice if latency is what we prioritize.

databroker/src/broker.rs
argerus: I'm saying:
> This should be forwarded to to the subscriber (of the broadcast channel),

i.e. to whoever called subscribe internally in databroker and is subscribing to this stream. That means the decision of what to do when it lags is located in the implementation of a specific service which can then decide whether to report that back over GRPC for example.

databroker/src/broker.rs
rafaeling: Done

databroker/src/broker.rs
rafaeling: Cant remember but it is fixed now.

databroker/src/broker.rs
rafaeling: Cant remember but it is fixed now.

databroker/src/broker.rs
rafaeling: No way found to propagate the lagged error to the `subscribe` caller, just found a way to pipe the error into the stream and return it back to the end user as tonic error.
The error is received while reading the `receiver`, no idea hot to stop the stream there and return an error to the `subscribe` caller, maybe a different function to 'filter_map' and 'map` can stop the stream and return an error?

argerus: I think you simply need to change the returned stream from `impl Stream<Item = EntryUpdates>` to `impl Stream<Item = Result<EntryUpdates>>` or something like that..

But perhaps it makes sense to postpone the implementation of that as we currently don't have the need to propagate the fact that a subscriber is lagging to the subscriber (i.e. the GRPC / VISS subscriber).

databroker/src/broker.rs
```diff
@@ -1622,12 +1630,14 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
-    ) -> Result<impl Stream<Item = EntryUpdates>, SubscriptionError> {
+        lag_buffer_capacity: Option<usize>,
+    ) -> Result<impl Stream<Item = Result<EntryUpdates, BroadcastStreamRecvError>>, SubscriptionError>
+    {
         if valid_entries.is_empty() {
             return Err(SubscriptionError::InvalidInput);
         }
 
-        let (sender, receiver) = mpsc::channel(10);
+        let (sender, receiver) = broadcast::channel(lag_buffer_capacity.unwrap_or(1));
```
argerus, line 1640: I don't think we should have the ability to set an unbounded (or 2^32) capacity here, as this is controlled by third party apps. That would probably be an easy denial of service vector.

databroker/src/broker.rs
argerus: > I suppose you mean only to add the size field parameter to `kuksa.val.v2` subscribe request, but what would be the appropriate default value to be used then for `kuksa.val.v1` and `sdv.databroker.v1`?

I actually didn't mean that you must add it to any GRPC interface right now (in this PR). I just said that it should be configurable in the (internal) call to subscribe, i.e. `pub async fn subscribe(...)`.

databroker/src/broker.rs
rafaeling: I can try your suggestion, anyway, have been trying hard to trigger the lagged error wit lazy subscriber but not way to produce it, never received on client side any lagged error, no idea why, maybe you can test it as well and see if you can get the corresponding error.

databroker/src/broker.rs
erikbosch: Good point. We do not yet have any config framework for Databroker where one can set limits like this, right? so for now just hard-code a sane max-limit (1000?) and give an error if exceeded.

(And give a hint in proto file, like "max value supported is implementation dependent")

rafaeling: Done

databroker/src/broker.rs
```diff
@@ -157,6 +161,7 @@ pub enum QueryError {
 pub enum SubscriptionError {
     NotFound,
     InvalidInput,
+    InvalidInputBroadChannel,
```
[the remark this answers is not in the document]
rafaeling: InvalidSubscibeBufferSize?

databroker/src/broker.rs
```diff
@@ -1622,12 +1633,23 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
     pub async fn subscribe(
         &self,
         valid_entries: HashMap<i32, HashSet<Field>>,
-    ) -> Result<impl Stream<Item = EntryUpdates>, SubscriptionError> {
+        lag_buffer_capacity: Option<usize>,
```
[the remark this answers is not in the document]
rafaeling: Done

databroker/src/broker.rs
```diff
@@ -1622,12 +1633,23 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+    ) -> Result<impl Stream<Item = Result<EntryUpdates, BroadcastStreamRecvError>>, SubscriptionError>
+    {
         if valid_entries.is_empty() {
             return Err(SubscriptionError::InvalidInput);
         }
 
-        let (sender, receiver) = mpsc::channel(10);
+        let channel_capacity = if let Some(cap) = lag_buffer_capacity {
+            if cap > MAX_BROADCAST_CHANNEL_CAPACITY {
```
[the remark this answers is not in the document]
rafaeling: Done

databroker/src/broker.rs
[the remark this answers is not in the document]
rafaeling: Done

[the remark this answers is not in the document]
argerus: Either just reuse `InvalidInput` or if adding a new one `BufferToLarge` or `InvalidBufferSize`.

databroker/src/broker.rs
```diff
@@ -157,6 +160,7 @@ pub enum QueryError {
 pub enum SubscriptionError {
     NotFound,
     InvalidInput,
+    InvalidSubscibeBufferSize,
```
argerus, line 163: There's a typo (Subscibe instead of Subscribe).

But might as well remove it completely and have `InvalidBufferSize` instead as "subscribe" implied by the fact it's a subscription error.

rafaeling: Updated

proto/kuksa/val/v2/val.proto
```diff
@@ -51,26 +51,39 @@ service VAL {
...
+  //       - if lag_buffer_capacity exceeds the maximum permitted
+  //             MAX_BROADCAST_CHANNEL_CAPACITY: usize = 1000;
   //
   // When subscribing, Databroker shall immediately return the value for all
   // subscribed entries.
   // If a value isn't available when subscribing to a it, it should return None
   //
+  // If a subscriber is slow to consume signals, messages will be buffered up
+  // to the specified lag_buffer_capacity before the oldest messages are dropped.
+  //
```
argerus, line 66:
```suggestion
  // If a subscriber is slow to consume signals, messages will be buffered up
  // to the specified buffer_size before the oldest messages are dropped.
  //
```

proto/kuksa/val/v2/val.proto
```diff
@@ -51,26 +51,39 @@ service VAL {
...
   // Returns (GRPC error code):
   //   NOT_FOUND if any of the signals are non-existant.
   //   UNAUTHENTICATED if no credentials provided or credentials has expired
   //   PERMISSION_DENIED if access is denied for any of the signals.
-  //   INVALID_ARGUMENT if the request is empty or provided path is too long
+  //   INVALID_ARGUMENT
+  //       - if the request is empty or provided path is too long
+  //             MAX_REQUEST_PATH_LENGTH: usize = 1000;
+  //       - if lag_buffer_capacity exceeds the maximum permitted
+  //             MAX_BROADCAST_CHANNEL_CAPACITY: usize = 1000;
```
argerus, line 78:
```suggestion
  //   INVALID_ARGUMENT
  //       - if the request is empty or provided path is too long
  //             MAX_REQUEST_PATH_LENGTH: 1000
  //       - if lag_buffer_capacity exceeds the maximum permitted
  //             MAX_BUFFER_SIZE: 1000
```

proto/kuksa/val/v2/val.proto
```diff
@@ -51,26 +51,39 @@ service VAL {
...
   //
   // When subscribing, Databroker shall immediately return the value for all
   // subscribed entries.
   // If a value isn't available when subscribing to a it, it should return None
   //
+  // If a subscriber is slow to consume signals, messages will be buffered up
+  // to the specified lag_buffer_capacity before the oldest messages are dropped.
+  //
```
argerus, line 86:
```suggestion
  // If a subscriber is slow to consume signals, messages will be buffered up
  // to the specified buffer_size before the oldest messages are dropped.
  //
```

proto/kuksa/val/v2/val.proto
```diff
@@ -191,14 +204,19 @@ message GetValuesResponse {
 
 message SubscribeRequest {
   repeated string signal_paths = 1;
+
+  // Specifies the number of messages that can be buffered for
+  // slow subscribers before the oldest messages are dropped.
+  uint32 lag_buffer_capacity   = 2;
```
argerus, line 210:
```suggestion

  // Specifies the number of messages that can be buffered for
  // slow subscribers before the oldest messages are dropped.
  uint32 buffer_size           = 2;
```

proto/kuksa/val/v2/val.proto
```diff
@@ -191,14 +204,19 @@ message GetValuesResponse {
...
 }
 
 message SubscribeResponse {
   map<string, Datapoint> entries = 1;
 }
 
 message SubscribeByIdRequest {
-  repeated int32 signal_ids = 1;
+  repeated int32 signal_ids  = 1;
+  uint32 lag_buffer_capacity = 2;
```
argerus, line 219:
```suggestion
  repeated int32 signal_ids  = 1;
  // Specifies the number of messages that can be buffered for
  // slow subscribers before the oldest messages are dropped.
  uint32 buffer_size         = 2;
```

argerus (changes requested):

PR #96, opened 2024-11-05
Remove of `optional` fields from sdv.databroker.v1 and kuksa.val.v2 due to older proto3 versions limitations.

rafaeling:


proto/sdv/databroker/v1/types.proto
```diff
@@ -174,20 +174,20 @@ message ValueRestriction {
...
+  uint64 min            = 1;
+  uint64 max            = 2;
   repeated uint64 allowed_values = 3;
 }
 
 message ValueRestrictionFloat {
-  optional double min = 1;
-  optional double max = 2;
+  double min = 1;
+  double max = 2;
```
argerus, line 190: This doesn't really work as numeric types in proto3 default to 0 (which would be indistinguishable from setting a min or max of 0).

I think the whole `ValueRestriction` concept is unnecessarily complex when compared to the previous approach of one field per restriction:
```proto
message Metadata {
  int32 id               = 1;
  EntryType entry_type   = 2;
  string name            = 4;
  DataType data_type     = 5;
  string description     = 7;

  Allowed allowed = 10;
  Value min = 11;
  Value max = 12;
};
```

So I would suggest that we revert adding `ValueRestriction` to sdv.databroker.v1, and preferrably update kuksa.val.v2 to use a more straight forward one field per restriction approach.

argerus (changes requested):
I would prefer if the kuksa.val.v1 concept of `ValueRestriction` was not carried over to the other protos

databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -267,8 +267,8 @@ fn value_restriction_from(metadata: &broker::Metadata) -> Option<proto::ValueRes
                     r#type: Some(proto::value_restriction::Type::Signed(
                         proto::ValueRestrictionInt {
                             allowed_values: allowed,
-                            min: min_value,
-                            max: max_value,
+                            min: min_value.unwrap_or(i64::MIN),
```
erikbosch, line 270: Will this not result in odd results for smaller VSS-types like `int8` where you for instance have `max: 53` but not any minimum defined. Will we then not return `i64::MIN` which is not correct and might be confusing for the client? 

I.e. do we need some helper method `get_implicit_min(metadata.data_type)` that returns for instance -128 for a `int8`

proto/kuksa/val/v2/types.proto
```diff
@@ -128,20 +128,20 @@ message ValueRestriction {
 }
 
 message ValueRestrictionInt {
-  optional sint64 min            = 1;
-  optional sint64 max            = 2;
+  sint64 min            = 1;
```
erikbosch, line 131: Maybe add some generic documentation/comment on that the values reported are the actual limits checked/enforced by Databroker, which may be either values specified in VSS, or if not present implicit limits for the datatype.

proto/sdv/databroker/v1/types.proto
erikbosch: But with the current implementation (in this PR) we always set min/max, right. I.e. if not specified in VSS we use for instance `f64::MIN`, so they are always explicitly set, right?

erikbosch:
I have nothing in particular against this PR, but do we know if this will have any impact on any existing clients. Any thoughts @lukasmittag ?

databroker/src/grpc/kuksa_val_v2/conversions.rs
```diff
@@ -363,11 +376,11 @@ impl From<&broker::Metadata> for proto::Metadata {
...
-            description: Some(metadata.description.clone()),
-            comment: None,
-            deprecation: None,
-            unit: metadata.unit.clone(),
-            value_restriction: value_restriction_from(metadata),
+            description: metadata.description.clone(),
+            comment: String::new(),
+            deprecation: String::new(),
+            unit: metadata.unit.clone().unwrap_or_default(),
+            value_restrictions: value_restriction_from(metadata),
```
lukasmittag, line 383: if you call this value_restrictions you should call the function the same
```suggestion
            value_restrictions: value_restrictions_from(metadata),
```

lukasmittag:
what about viss parts?

proto/sdv/databroker/v1/types.proto
```diff
@@ -144,56 +144,52 @@ message Datapoint {
...
-  // int32             max_update_hz       = 11; // Only for CONTINUOUS
+  int32 id                             = 1;
+  EntryType entry_type                 = 2;
+  string name                          = 4;
+  DataType data_type                   = 5;
+  ChangeType change_type               = 6;  // CONTINUOUS or STATIC or ON_CHANGE
+  string description                   = 7;
+
+  // Value restrictions
+  ValueRestrictions value_restrictions = 10;
```
erikbosch, line 155: Do we know if there are anyone actually using ValueRestriction in this old API?
Because this will be a breaking change for anyone actually using this, right?

erikbosch:
We should better add in the PR description or commit description why we remove optional. If I remember correct it is because older proto3 versions does not support it, right?, but if so we should better document it so we do not forget it. Possibly even worth mentioning in proto to avoid adding new optional fields later.

argerus:
> Possibly even worth mentioning in proto to avoid adding new optional fields later.

I think that's a very good idea.

databroker/src/grpc/kuksa_val_v2/conversions.rs
rafaeling: This was removed and `None` placed instead, so it should be resolved

proto/kuksa/val/v2/types.proto
rafaeling: Added

rafaeling:
> We should better add in the PR description or commit description why we remove optional. If I remember correct it is because older proto3 versions does not support it, right?, but if so we should better document it so we do not forget it. Possibly even worth mentioning in proto to avoid adding new optional fields later.

Added

proto/sdv/databroker/v1/types.proto
```diff
@@ -144,56 +145,54 @@ message Datapoint {
...
+}
+
+// Restriction fields are checked/enforced by Databroker.
+// They can be either values specified in VSS, or if not present implicit limits for the datatype.
+message ValueRestrictions {
+  Allowed allowed      = 1;
+  ValueRestriction min = 2;
+  ValueRestriction max = 3;
+}
```
argerus, line 165: I don't see why we need to create an additional message to contain the value restrictions. All the other metadata is at the "root" level which IMO makes it easier to work with.

I think this should be reverted to what was already there, i.e.

```
message Metadata {
  ..
  Allowed allowed = 10;
}
```

and support for mix/max can be added along these lines:
```rust
message Metadata {
  ..
  Allowed allowed      = 10;
  ValueRestriction min = 11;
  ValueRestriction max = 12;
}
```

argerus (changes requested):

proto/sdv/databroker/v1/types.proto
rafaeling: First thing I tried was your suggestion, but I decided to wrap all in same message since the three messages are somehow related with each other to restrict user's update values and also the way it looks now the conversion call is more (imo) cleaner and easy to understand with just one message field (see-> https://github.com/eclipse-kuksa/kuksa-databroker/pull/96/files#diff-d482c91dd456ce51fbebd32a722dab7e339bfda360b1ab807c296ba95ca18179R383)

proto/kuksa/val/v2/types.proto
```diff
@@ -74,82 +75,58 @@ enum ErrorCode {
...
-    ValueRestrictionUint unsigned        = 23;
-    // For floating point VSS values (float and double)
-    ValueRestrictionFloat floating_point = 24;
-  }
+// Restriction fields are checked/enforced by Databroker.
+// They can be either values specified in VSS, or if not present implicit limits for the datatype.
+message ValueRestrictions {
+  Allowed allowed = 1;
+  Value min       = 2;
+  Value max       = 3;
```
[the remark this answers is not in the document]
rafaeling: For me is hard to see as a good alternative `Value allowed_values` if only array types are needed for `allowed` field.
Would it be actually better to have something like this since it makes no sense imo to have allowed and min/max fields at same time?

```proto
message Metadata {
  ..
  // Value restrictions checked/enforced by Databroker
  ValueRestrictions value_restrictions = 1;
}

message ValueRestrictions {
  oneof value {
    Allowed allowed                       = 1;
    MinMaxRestriction min_max_restriction = 2;
  }
}

message MinMaxRestriction {
  Value min = 1;
  Value max = 2;
}
```

proto/sdv/databroker/v1/types.proto
argerus: I don't really see how putting these inside a `ValueRestriction` message makes it cleaner / easier to understand.

It seems to be the only field that requires the whole metadata struct in order to be constructed. And to construct it, first `metadata.allowed`, `metadata.min` and `metadata.max` are checked individually inside `value_restriction_from(...)` and if one of them is set a `ValueRestriction` is created:

```rust
fn value_restriction_from(metadata: &broker::Metadata) -> Option<proto::ValueRestrictions> {
  ...
  if min_value.is_some() | max_value.is_some() | allowed.is_some() {
     return Some(proto::ValueRestrictions { allowed: ..., min: ..., max: ...})
  }
  ...
}
```

How is that cleaner / easier to understand than how all the other metadata is converted, i.e. something like this:
```rust
proto::Metadata {
  allowed_values: metadata.allowed.into(),
  min: metadata.min.into(),
  max: metadata.max.into(),

}
```

proto/kuksa/val/v2/types.proto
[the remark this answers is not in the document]
argerus: It's the same as `Value min` and `Value max` being able to contain an array type. That is also not allowed, but enforced by databroker instead of at the proto level.

And it's the same thing everywhere a `Value` is used => it's up to databroker to enforce that the `Value` actually contains what is logically allowed (i.e. by checking the `DataType` of the entry.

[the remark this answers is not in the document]
argerus: And the suggested `Allowed` proto message doesn't enforce that the array type matches the datatype of the entry, so databroker has to enforce that manually anyway.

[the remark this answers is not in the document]
erikbosch: I have no strong opinion, but sort of like the simplicity of the suggestion from John:

```
message Metadata {
  ..
  // Value restrictions checked/enforced by Databroker
  Value allowed_values = 17; // Must be of array type
  Value min            = 18;
  Value max            = 19;
}
```

proto/sdv/databroker/v1/types.proto
rafaeling: Done

proto/kuksa/val/v2/types.proto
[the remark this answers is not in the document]
rafaeling: Done

argerus (approved):

PR #97, opened 2024-11-05
"Add 'otel' feature for OpenTelemetry support

LikhithST:
Enabling the 'otel' feature conditionally includes OpenTelemetry and related tracing functionalities, allowing method-level tracing and context propagation."


databroker/src/broker.rs
```diff
@@ -579,7 +593,10 @@ impl Entry {
             self.actuator_target = actuator_target;
             changed.insert(Field::ActuatorTarget);
         }
-
+        if let Some(metadata_description) = update.description {
+            self.metadata.description = metadata_description;
+            // changed.insert(Field::ActuatorTarget);
+        }
```
lukasmittag, line 599: these changes at description are not mergeable we need to find another way to propagate the context to the client/provider.

databroker/src/broker.rs
```diff
@@ -728,6 +787,19 @@ impl ChangeSubscription {
...
+                                            {
+                                            let mut metadata = MetadataMap::new();
+                                            //  @TODO: Speak to Kuksa team regarding MetadataMap in proto file
+                                            let mut injector = MetadataMapInjector(&mut metadata);
+                                            opentelemetry::global::get_text_map_propagator(|propagator| {
+                                                propagator.inject_context(&current_span.context(), &mut injector);
+                                            });
+                                            let description = metadatamap_to_string(&metadata);
+
+                                            update.description = Some(description);
```
lukasmittag, line 801: please do not missuse description. Description is something to describe a datapoint.

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -976,13 +1049,24 @@ impl broker::EntryUpdate {
         } else {
             None
         };
+        let metadata_description = if fields.contains(&proto::Field::MetadataDescription) {
```
lukasmittag, line 1052: same here do not missuse description please.

databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -976,13 +1049,24 @@ impl broker::EntryUpdate {
...
+            None
+        };
         Self {
             path: None,
             datapoint,
             actuator_target,
             entry_type: None,
             data_type: None,
-            description: None,
+            description: metadata_description.cloned(),
```
lukasmittag, line 1069: here as well

lukasmittag (changes requested):
update the branch pls

databroker/src/broker.rs
LikhithST: I was thinking if we could add this to the `types.proto` file

```
message MetadataMap {
  map<string, string> headers = 1;
}

message Metadata {
  optional MetadataMap otel_context = 18;  // to add this field alongside already existing fields in Metadata
}
```
By doing this, a new field is added to Metadata alongside description, while allowing the description to continue functioning as before _( I will revert all changes made to the description, ensuring it functions as it originally did )_.

MetadataMap could be used for otel context propagation and here is the output of `subscribe call` with the field `otel_context`: 

```
{
    "updates": [
        {
            "fields": [
                "FIELD_VALUE"
            ],
            "entry": {
                "path": "Vehicle.Speed",
                "value": {
                    "timestamp": {
                        "seconds": "1733672625",
                        "nanos": 127403578
                    },
                    "float": 22
                },
                "actuator_target": null,
                "metadata": {
                    "data_type": "DATA_TYPE_UNSPECIFIED",
                    "entry_type": "ENTRY_TYPE_UNSPECIFIED",
                    "value_restriction": null,
                    "description": "Odometer reading, total distance traveled during the lifetime of the vehicle.",
                    "unit": "km/h",
                    "otel_context": {
                        "headers": {
                            "tracestate": "",
                            "traceparent": "00-d9b7e6227856aefc7fe4ba0e142e2017-b6923a717115ffae-01"
                        }
                    }
                }
            }
        }
    ]
}
```

I have tested this change locally and it works as expected.

Please let me know if this approach is acceptable.

lukasmittag: no changes in proto as of now. Could you confirm that the tracing stuff is working with the context stuff removed?

lukasmittag:
no documentation how to start etc. would be nice if there would be something. As well as an example of a client & provider example for context propagation.

databroker/src/broker.rs
LikhithST: Yes, I have created a new pull request #119 containing just the tracing functionality. 
I have tested this change locally and it works fine.

LikhithST:
I will add readme explaining environment variables required and context propagation mechanism between the clients and the databroker in detail in this branch testing-main.

mikehaller:
What would be the benefit of having the OTel Context available? I guess the traces would be more "complete", e.g. spanning across multiple services also outside of the databroker tracing. Is my assumption correct?

lukasmittag:
@LikhithST this can be closed since #92 is merged?

LikhithST:
Hi @lukasmittag, 
If it's not planned to include context propagation for `otel` feature in the near future, We can close this PR.

PR #99, opened 2024-11-11
Improved throughput: reduce unnecessary overload on gRPC bidirectional stream by avoiding empty responses

rafaeling:
Databroker was underperforming compared to `sdv.databroker.v1` APIs implementation.
gRPC bidirectional stream buffer was being overloaded due to the unnecessary return of empty responses, which caused reduced throughput and buffer congestion.

### DIFF
--- a/databroker/src/grpc/kuksa_val_v2/val.rs
+++ b/databroker/src/grpc/kuksa_val_v2/val.rs
@@ -2212,18 +2212,18 @@ mod tests {
                 let entries_size = list_response.metadata.len();
                 assert_eq!(entries_size, 1);
 
-                let value_restriction = Some(proto::ValueRestriction {
-                    r#type: Some(proto::value_restriction::Type::Signed(
-                        proto::ValueRestrictionInt {
-                            allowed_values: Vec::new(),
-                            min: Some(-7),
-                            max: Some(19),
-                        },
-                    )),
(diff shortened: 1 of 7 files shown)


databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -646,9 +646,10 @@ impl proto::val_server::Val for broker::DataBroker {
                                             },
                                             Some(PublishValuesRequest(publish_values_request)) => {
                                                 let response = publish_values(&broker, &publish_values_request).await;
-                                                if let Err(err) = response_stream_sender.send(Ok(response)).await
-                                                {
-                                                    debug!("Failed to send response: {}", err);
+                                                if let Some(value) = response {
```
erikbosch, line 649: We should better clarify in proto files when response shall be sent. Today my interpretation of the text below is that you ALWAYS should get a Response message. We should better clarify with something like "_Databroker returns PublishValuesResponse upon error, and nothing upon success_"

Current text:

 _Provider sends PublishValuesRequest -> Databroker returns PublishValuesResponse_

Similar for other similar cases

rafaeling:
I close this one and create a new one, branched off the wrong branch. https://github.com/eclipse-kuksa/kuksa-databroker/pull/100

PR #100, opened 2024-11-11
Improved throughput: reduce unnecessary overload on gRPC bidirectional stream by avoiding empty responses

rafaeling:
Databroker was underperforming compared to sdv.databroker.v1 APIs implementation.
gRPC bidirectional stream buffer was being overloaded due to the unnecessary return of empty responses, which caused reduced throughput and buffer congestion.

**If we accept this change, Provider will fire and forget since Databroker will not send back any ACK message. Could it overload the gRPC buffer?**

### DIFF
--- a/databroker/src/grpc/kuksa_val_v2/val.rs
+++ b/databroker/src/grpc/kuksa_val_v2/val.rs
@@ -648,7 +648,7 @@ impl proto::val_server::Val for broker::DataBroker {
                                                 let response = publish_values(&broker, &publish_values_request).await;
                                                 if let Some(value) = response {
                                                     if let Err(err) = response_stream_sender.send(Ok(value)).await {
-                                                        debug!("Failed to send response: {}", err);
+                                                        debug!("Failed to send error response: {}", err);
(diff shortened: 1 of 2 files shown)


databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -793,17 +794,8 @@ async fn publish_values(
 
     // TODO check if provider is allowed to update the entries for the provided signals?
```
erikbosch, line 795: Noticed the TODO - Is it so that we currently do not check token, even if broker is configured to require token?
If so, something we better should fix before release - not necessarily part of this PR?

argerus (approved):

databroker/src/grpc/kuksa_val_v2/val.rs
erikbosch: But we have `let broker = broker.authorized_access(&permissions);` earlier, so it is "broker" that does the check on the line below. Could you confirm @argerus

PR #104, opened 2024-11-15
Accept 0 as buffer size

erikbosch:
I noticed that we did not support the implicit default value 0 for `buffer_size`, i.e. any user would need to specify buffer size, and I do not really see that as something that provide any extra value, so proposing a change in behavior.

- Make 0 an allowed value
- Change the meaning so that `buffer_size` indicates number of "slots" for older messages, i.e. if you specify `1` you will actually use `2 ` in code.

With this change the code can be simplified, as we do not need the special check for `buffer_size==0`.

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -1642,8 +1642,10 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
             if cap > MAX_SUBSCRIBE_BUFFER_SIZE {
                 return Err(SubscriptionError::InvalidBufferSize);
             }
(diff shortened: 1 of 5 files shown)


argerus (changes requested):
The call sites to `broker::subscribe(...)` should be changed to have `buffer_size: None` in order to use the default value (or `buffer_size: Some(0)` to be explicit)..

erikbosch:
Now using default (None) in "old" APIs. Also updated test cases to check limits

argerus (approved):

PR #106, opened 2024-11-20
Changing ListMetaData to correspond to reality

erikbosch:
We do not use the `filter` field, and the name `root` is not that logical. In older API supporting same wildcard syntax we use the term `path`, so proposing to change to that.

https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/proto/kuksa/val/v1/val.proto#L104

```
message SubscribeEntry {
  string path           = 1;
  View view             = 2;
  repeated Field fields = 3;
}
```

Related to discussion at https://github.com/eclipse-kuksa/kuksa-databroker/pull/105/files#r1848167147

### DIFF
--- a/proto/kuksa/val/v2/val.proto
+++ b/proto/kuksa/val/v2/val.proto
@@ -128,7 +128,7 @@ service VAL {
   // List metadata of signals matching the request.
   //
   // Returns (GRPC error code):
-  //   NOT_FOUND if the specified root branch does not exist.
+  //   NOT_FOUND if no signals matching the request are found.
   //   UNAUTHENTICATED if no credentials provided or credentials has expired
   //   INVALID_ARGUMENT if the provided path or wildcard is wrong.
   //
@@ -247,8 +247,9 @@ message BatchActuateResponse {
 }
 
 message ListMetadataRequest {
-  string root   = 1;
-  string filter = 2;
+  // A path which may include wildcards
+  // See https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/doc/wildcard_matching.md
+  string path   = 1;
 }
 
 message ListMetadataResponse {
--- a/databroker/src/grpc/kuksa_val_v2/val.rs
+++ b/databroker/src/grpc/kuksa_val_v2/val.rs
@@ -479,7 +479,7 @@ impl proto::val_server::Val for broker::DataBroker {
     }
 
     // Returns (GRPC error code):
-    //   NOT_FOUND if the specified root branch does not exist.
+    //   NOT_FOUND if no signals matching the request are found.
     //   UNAUTHENTICATED if no credentials provided or credentials has expired
     //   INVALID_ARGUMENT if the provided path or wildcard is wrong.
     //
@@ -499,7 +499,7 @@ impl proto::val_server::Val for broker::DataBroker {
 
         let metadata_request = request.into_inner();
 
-        match Matcher::new(&metadata_request.root) {
+        match Matcher::new(&metadata_request.path) {
             Ok(matcher) => {
                 let mut metadata_response = Vec::new();
                 broker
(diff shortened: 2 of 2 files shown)


erikbosch:
There is ongoing discussion if we better should change implementation to reflect the *.proto file. I.e. have

- root - that must be a path to branch without wildcards. 
  - OK Examples `Vehicle`, `Vehicle.Cabin`.
  - Bad Examples: `Vehicle*`, `Vehicle.Speed`, `*`
- filter: Path possibly including wildcard

Below some possible examples to show results for a spec containing 

```
Branch A
  Branch A.B
    Sensor A.B.C
    Sensor A.B.CC
    Sensor A.B.D
  Branch A.BB
    Sensor A.BB.A
  Branch A.C // Empty branch
```


Root | Filter | Result
-----|------|--
`A`| `None` | `A.B.C`, `A.B.CC`,  `A.B.D` and `A.BB.A`
`A`| `""` | `A.B.C`, `A.B.CC`,  `A.B.D` and `A.BB.A`
`A`| `*` | `A.B.C`, `A.B.CC`,  `A.B.D` and `A.BB.A` **Or should `*` only match a single level to differentatiate from `**`**
`A.B`| `None` | `A.B.C`, `A.B.CC` and `A.B.D`
`A.B`| `""` | `A.B.C`, `A.B.CC` and `A.B.D`
`A.B`| `*` | `A.B.C`, `A.B.CC` and `A.B.D`
`A.B`| `C` | `A.B.C` Only!
`A.B`| `C*` | `A.B.C`  and `A.B.CC`
`A.B*`| `None` | Error - wildcard not allowed in root - `INVALID_ARGUMENT`
`A.B.C`| `None` | Error - root must not be a signal!  - `INVALID_ARGUMENT`
`A.C`| `None` | Empty list returned as branch exist but is empty (_if we can differentiate the case from the next_)
`A.D`| `None` | Error - Branch does not exist  - `NOT_FOUND`
`A.B`| `Sebastian` | Empty list returned as signal `A.B.Sebastian` does not exist


FYI @rafaeling 

@argerus - does the table match how you think it should work

erikbosch:
Put on hold for now, we possibly want to have root plus filter anyway, so at least no proto update for 5.0

proto/kuksa/val/v2/val.proto
```diff
@@ -247,8 +247,9 @@ message BatchActuateResponse {
 }
 
 message ListMetadataRequest {
-  string root   = 1;
-  string filter = 2;
+  // A path which may include wildcards
+  // See https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/doc/wildcard_matching.md
+  string path   = 1;
```
BjoernAtBosch, line 252: Some points of critic:
* With the removed filter one cannot request certain "types" of metadata any longer (like "I just want to get the numeric ids").
* Having just one path (formerly root), it's hard to request metadata in a single request for completely different branches, like `Vehicle.Speed` and `Vehicle.Cabin.Seat.*`. (Or?)

proto/kuksa/val/v2/val.proto
BjoernAtBosch, line 1: (Don't know, where to put this finding else:)
Rename field `datapoints` of `PublishValuesRequest` to `data_points`(w/ underscore).

proto/kuksa/val/v2/val.proto
erikbosch: Well as of today we do not care about "filter" at all. I have no problem having a filter field if we actually intend to use it, but we do not describe how filter shall/can be used at all right now, like if it going to support the use-cases you mention. But as we seem to be uncertain I put this PR on hold. 

In #105 I instead pointed out that `filter` currently is not used, and that current implementation may change. Like to be on the safe side, do not use wildcards in root, as it may not be supported in the future.

proto/kuksa/val/v2/val.proto
erikbosch: Fixed by #108

SebastianSchildt:
This looks a bit outdated, but I do see quite a bit of discussion. Should I close it for now, or is there still work being done?

@erikbosch  @BjoernAtBosch  @rafaeling  ?

lukasmittag:
outdated- closing this since it seems not relevant anymore.

PR #108, opened 2024-11-25
Rename datapoints

erikbosch:
Based on a comment at https://github.com/eclipse-kuksa/kuksa-databroker/pull/106#discussion_r1853649367 from @BjoernAtBosch 

Should we possibly use `data_points` instead of `datapoints`. Most other members in v2 use the form `data_point` / `data_points`

If we are to change it, maybe it is good to change it now.


lukasmittag (approved):
LGTM

PR #110, opened 2024-11-26
Update version to 0.6.0-dev.0

erikbosch:
Performed by `scripts/prepare_release.sh 0.6.0-dev.0`. This is to make sure that if someone builds from source it should be clearly visible that they are not using the official 0.5.0 version.


rafaeling (approved):
LGTM

PR #111, opened 2024-11-26
Change release flow

erikbosch:
This is a proposal to change the release handling somewhat

# Today 

* Create a tag and push to remote
* Trigger workflow on the tag and mention wanted version (e.g. `1.2.3`)

*If you trigger on branch you will get a faulty tag proposal in draft release that you must correct, tag will be created when release is published*

# After this PR

* Create a tag and push to remote
* Creation of release is started automatically

*If you run the workflow on a branch you will get a release named after the branch, which you likely do not want*

### DIFF
--- a/.github/workflows/create_draft_release.yml
+++ b/.github/workflows/create_draft_release.yml
@@ -14,17 +14,11 @@
 name: Create Draft Release
 
 on:
-  workflow_dispatch: # input version manually. Overrides push tag
-    inputs:
-      tag:
-        description: "Release version, eg:latest, 0.2.1"
-        required: true
-        default: "0.0.0"
+  workflow_dispatch: # select tag when creating
+  push:
+   tags:
+     - "*.*.*"
 
-  # As of today trigger only manually
-  #push:
-  #  tags:
-  #    - "*.*.*"
 
 # Needed if GITHUB_TOKEN by default do not have right to create release
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/create_draft_release.yml
```diff
@@ -14,17 +14,11 @@
 name: Create Draft Release
 
 on:
-  workflow_dispatch: # input version manually. Overrides push tag
-    inputs:
-      tag:
-        description: "Release version, eg:latest, 0.2.1"
-        required: true
-        default: "0.0.0"
+  workflow_dispatch: # select tag when creating
```
lukasmittag, line 17:
```suggestion
  on:
     release:
         types: [published]
```

use this instead?

erikbosch: What would be the intended use case? We have an action further down in the script ( `softprops/action-gh-release@v2`) to create a release, and that becomes quite useless if we already have created a release.

And adding `on:release` is a bit superfluous. If you create a new release, and the tag does not exist since before a new tag will be created and that will anyway trigger this workflow.

I am actually considering if we should remove `workflow_dispatch` - I do not see that many use-cases for it, except possibly if you have yanked a release and want to rebuild a new release for the same tag.

SebastianSchildt (approved):
Looks useful. I think the risk is sort of low, it creates draft releases, that can likely be deleted via GH WebUI if really needed.

PR #112, opened 2024-11-29
Fix ListMetadata

erikbosch:
This intends to make the implementation consistent with the documentation.

In short, Databroker does not know if A.B refer to a signal A.B or a branch A.B. We solve that by both looking for A.B.** and A.B.
That is OK as if there is a signal A.B there cannot be a branch A.B, even if it may require somewhat more comparisons, but you cannot do it in one comparision using glob.

Also our current glob cargo package does not seem to be maintained, we may consider migrating to something else.

How to test:

Run the three commands below on latest released version as well as with this PR

```

# This one previously failed as we did not add .**

docker run --network=host fullstorydev/grpcurl  -d '{ "root": "Vehicle.Cabin.Sunroof" }' -plaintext localhost:55555 kuksa.val.v2.VAL.ListMetadata

# This one worked before as we added .** if path did not contain dots

docker run --network=host fullstorydev/grpcurl  -d '{ "root": "Vehicle" }' -plaintext localhost:55555 kuksa.val.v2.VAL.ListMetadata

# This one worked before as it reference a signal

docker run --network=host fullstorydev/grpcurl  -d '{ "root": "Vehicle.Speed" }' -plaintext localhost:55555 kuksa.val.v2.VAL.ListMetadata
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ https://www.eclipse.org/projects/handbook/#resources-commit
 
 Contact the project developers via the project's "dev" list.
 
-* https://dev.eclipse.org/mailman/listinfo/kuksa-dev
+* https://accounts.eclipse.org/mailing-list/kuksa-dev
 
 ## Pre-commit set up
 This repository is set up to use [pre-commit](https://pre-commit.com/) hooks.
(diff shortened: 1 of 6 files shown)


SebastianSchildt (approved):
lgtm 🍏

PR #113, opened 2024-12-05
Build for profile release-with-debug

mikehaller:
A build workflow for the `release-with-debug` profile, which creates a databroker binary with debug symbols enabled.
This is useful for the integration into an eBPF-based performance test environments.

As separate testers would need to rebuild databroker with that profile, which requires them to set up their own databroker development environment, we've decided for convenience to release debug versions of databroker for exactly that use case.

The additional `debugsymbols.txt` file in the archive contains the entry point for eBPF probes to instrument the function within databroker where the VSS Signals are being updated by providers:

```
DatabaseWriteAccess.update()
```

Also added some DevContainer setup and more documentation, e.g. about the Slow-Subscriber example, which can be used to test for lagging updates.


scripts/build-databroker.sh
```diff
@@ -112,12 +123,13 @@ function build_target() {
     #
     # this is solved by using different target-dirs for each platform
     echo "Building databroker for target $target_rust"
-    cross build --target $target_rust --target-dir ./target-$target_docker --features $KUKSA_DATABROKER_FEATURES --bin databroker --release
+    env
+    cross build --target $target_rust --target-dir ./target-$target_docker --features $KUKSA_DATABROKER_FEATURES --bin databroker --profile ${KUKSA_DATABROKER_PROFILE}
```
erikbosch, line 127: Is `--release` and `--profile release` equivalent ways to express the same thing?

.github/workflows/debug_build.yml
```diff
@@ -0,0 +1,357 @@
...
+# * information regarding copyright ownership.
+# *
+# * This program and the accompanying materials are made available under the
+# * terms of the Apache License 2.0 which is available at
+# * http://www.apache.org/licenses/LICENSE-2.0
+# *
+# * SPDX-License-Identifier: Apache-2.0
+# ********************************************************************************/
+
+name: Build kuksa-databroker
```
erikbosch, line 14: Change name to something else to avoid confusion in UI, e.g. just add `(debug)` at end

.github/workflows/debug_build.yml
erikbosch, line 1: Long term if we want multiple variants (like debug/release) we should possibly add that as an optional flag to the existing workflow and have wrappers that calls the wanted variant, similar to how https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/.github/workflows/create_draft_release.yml call builds, to avoid duplication.

scripts/build-databroker.sh
lukasmittag: have only tried `--profile release` yet

.github/workflows/debug_build.yml
```diff
@@ -0,0 +1,357 @@
...
+    secrets:
+      QUAY_IO_TOKEN:
+        required: true
+      QUAY_IO_USERNAME:
+        required: true
+  workflow_dispatch:
+
+# suffix to avoid cancellation when running from release workflow
+concurrency:
+      group: ${{ github.ref }}-${{ github.workflow }}-databroker
```
erikbosch, line 30: Add `-debug` to the end of this line, i.e. `-databroker-debug`

Could possibly save use some problem if we would build a debug variant for every release as well.

.devcontainer/Dockerfile
erikbosch, line 1: Before we merge this we should better add something in documentation on how to use it, and after merge also something in https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Testing on how to test it.

As an example - if you run Databroker in a devcontainer, what do you need do in devcontainer.json to make the port available?

databroker/databroker_bin.cdx.json
```diff
@@ -0,0 +1,9900 @@
+{
```
erikbosch, line 1: Intentional to commit Cyclone dx files? If so - how/when shall they be updated? Should possibly be documented. As they mention Databroker version I assume they need to be updated after you have updated version according to https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Process#update-rust-versions and possibly

FYI @SebastianSchildt

.github/workflows/debug_build.yml
mikehaller: For now, they are slightly different (e.g. a debug build doesn't need to redo the compliance check), hence it's a separate workflow.

.github/workflows/debug_build.yml
mikehaller: Thx, done

.devcontainer/Dockerfile
mikehaller: Good idea, thank you

databroker/databroker_bin.cdx.json
mikehaller: Thank you, removed the unintentionally committed CycloneDX files.

SebastianSchildt:
Pls also check pre-commit line break issues https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/CONTRIBUTING.md#pre-commit-set-up

mikehaller:
> I noticed there are some .devcontainer configs not previously there. If they are not part/required for this feature, maybe this is something for a separate PR?

Moved to separate #126, #127 and #128

PR #115, opened 2024-12-06
Use samllvec to reduce heap use, use inline functions and general improvements

rafaeling:
Improved performance by around 6%

### DIFF
--- a/databroker/Cargo.toml
+++ b/databroker/Cargo.toml
@@ -62,6 +62,7 @@ lazy_static = "1.4.0"
 thiserror = "1.0.47"
 futures = { version = "0.3.28" }
 async-trait = "0.1.82"
+smallvec = "1.13.2"
 
 # VISS
 axum = { version = "0.6.20", optional = true, features = ["ws"] }
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -14,6 +14,8 @@
 use crate::permissions::{PermissionError, Permissions};
 pub use crate::types;
 
+use smallvec::SmallVec;
+
 use crate::query;
 pub use crate::types::{ChangeType, DataType, DataValue, EntryType};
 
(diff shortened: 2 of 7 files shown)


databroker/Cargo.toml
```diff
@@ -62,6 +62,7 @@ lazy_static = "1.4.0"
 thiserror = "1.0.47"
 futures = { version = "0.3.28" }
 async-trait = "0.1.82"
+smallvec = "1.13.2"
```
erikbosch, line 65: I assume you should update the lock file as well, or?

databroker/src/broker.rs
```diff
@@ -112,6 +114,10 @@ pub enum Field {
     ActuatorTarget,
     MetadataUnit,
 }
+#[derive(Debug, Clone)]
+pub struct StackVecField {
+    pub svec: SmallVec<[Field; 3]>,
```
lukasmittag, line 119: do we expect 3 to be the "optimal" number because if I understand correctly after 3 elements it will store it anyways in the heap right?

lukasmittag: for ` pub fields: StackVecField,` this might be the right approach since fields is [current, target, metadata]

databroker/src/broker.rs
```diff
@@ -1493,6 +1567,15 @@ impl AuthorizedAccess<'_, '_> {
             .cloned()
     }
 
+    pub async fn contains_id(&self, id: i32) -> bool {
```
lukasmittag, line 1570: same name as function in line 1263 intended?

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -849,6 +847,7 @@ async fn publish_values(
     }
 }
 
+#[inline]
```
lukasmittag, line 850: is this function not a bit too long for inline?

lukasmittag:
yes lets close it and link it in an issue? @rafaeling

PR #116, opened 2024-12-07
Log a warning for slow signal updates to subscribers

mikehaller:
When subscribers are too slow to process vehicle signal updates, databroker will log a warning.

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -1670,10 +1670,10 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
             .await
             .add_change_subscription(subscription);
 
-        let stream = BroadcastStream::new(receiver).filter_map(|result| match result {
+        let stream = BroadcastStream::new(receiver).filter_map(move |result| match result {
             Ok(message) => Some(message),
             Err(err) => {
-                debug!("Lagged entries: {}", err);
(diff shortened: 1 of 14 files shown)


databroker/src/broker.rs
```diff
@@ -1203,7 +1203,7 @@ impl<'a, 'b> DatabaseReadAccess<'a, 'b> {
     }
 }
 
-impl<'a, 'b> DatabaseWriteAccess<'a, 'b> {
+impl DatabaseWriteAccess<'_, '_> {
```
rafaeling, line 1206: Why did you changed from explicit lifetimes to anonymous lifetimes? did it complain?

erikbosch: This is inherited from #114, see also https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes

As we use "latest and greatest" clippy we get these types of regressions now and then

(long term we could discuss if we rather should have a fixed rust/clippy/fmt version and perform updates only when preparing for a (major) release, to reduce the risk of regressions)

lukasmittag:
whats the state here still a draft. Intended?

SebastianSchildt:
This looks a bit abandoned, and is not even targetting current main. If there are not vetos @mikehaller @lukasmittag  I would close it for now.

Of course equally happy if someone invests some time into this, but currently it is a zombie :D

mikehaller:
Please close

PR #117, opened 2024-12-10
Minor changes to info and error messages for subscribe & provide_actuation

lukasmittag:
The now given error is to undetailed it just says "Invalid argument". The error can just originate if you do not give any valid path (or id in v2 use case).
per sé the provide_actuation is a subscription as well but we should to the outside world communicate/use a more detailed info message if the provider goes away. For example removing provided actuation.


wba2hi (approved):
LGTM

PR #118, opened 2024-12-11
Fix no-tls build for databroker

AkhilTThomas:
This PR fixes the broken build when building `kuksa-databroker` with `--no-default-features` flag.
Additionally i have also enforced that `tls` feature flag for `kuksa` and `common` are not used in `no-default-feature` mode.

### DIFF
--- a/databroker-cli/Cargo.toml
+++ b/databroker-cli/Cargo.toml
@@ -48,4 +48,4 @@ http = "0.2.8"
 
 [features]
 default = ["tls"]
-tls = ["tonic/tls"]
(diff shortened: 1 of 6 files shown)


SebastianSchildt (approved):
Looking good. I verified TLS connection is still in the default builds from CI and works.

AkhilTThomas:
> Looking good. I verified TLS connection is still in the default builds from CI and works.

It would be good if `-no-defaut-features` build is added to CI to ensure that this is caught sooner

SebastianSchildt:
Agree, I created a seperate issue for that (PRs welcome of course :D ) to not blocking the immediate fix

Issue #120, opened 2024-12-12
Add no-default feature build in CI


SebastianSchildt:
Making sure the no-default feature (currently mainly disabling TLS support) is build in CI, to detect breakage

See #118

SebastianSchildt:
Done

PR #121, opened 2024-12-13
Add cargo check for all features

AkhilTThomas:
This PR adds cargo-hack to build each target with all the possible feature flags. 
Addresses #120

### DIFF
--- a/.github/workflows/kuksa_databroker_build.yml
+++ b/.github/workflows/kuksa_databroker_build.yml
@@ -160,6 +160,9 @@ jobs:
       # Needed for pip
       with:
         python-version: '3.12'
+    - uses: taiki-e/install-action@v2.9.4
+      with:
+        tool: cargo-hack@0.6
     - name: Install build prerequisites
       working-directory: ${{github.workspace}}/
       run: |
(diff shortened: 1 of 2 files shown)


AkhilTThomas:
Caught an [issue](https://github.com/eclipse-kuksa/kuksa-databroker/actions/runs/12320971306/job/34391391578?pr=121) in `databroker-cli` for `--no-default-features` 
So it should address #120

SebastianSchildt:
@CodiumAI-Agent /review

.github/workflows/kuksa_databroker_build.yml
```diff
@@ -173,6 +176,10 @@ jobs:
       run: |
         ./scripts/build-databroker.sh ${{ matrix.platform.name }}
 
+    - name: "Check each feature"
+      working-directory: ${{github.workspace}}
+      run: cargo hack check --each-feature
```
SebastianSchildt, line 181: It seems we run this for each platform in the build matrix, but I think we do nothing platform specific here, so maybe moving it out can save some time?

SebastianSchildt: Maybe in the (unit) test job?

AkhilTThomas: @SebastianSchildt This adds ~5m 12s runtime for the unit test action.

SebastianSchildt (approved):
Nice. Looking good. As for the runtime, this should also profit from the caching. We can come back if it gets too long/annoying

🏖️

PR #123, opened 2024-12-30
Handle `None` for signal_id in GetValue endpoint

engelmi:
### Description of bug and proposed fix

Currently, calling the GetValue endpoint with an empty body will cause the databroker to panic since unwrap is called on the signal_id option - which is None in that case. In order to prevent the server-side panic, a simple none check is performed before calling unwrap.

### How to reproduce the bug

Example script to trigger the databroker to panic:
```python
import grpc

from kuksa.val.v2.val_pb2 import GetValueRequest
from kuksa.val.v2.val_pb2_grpc import VALStub

with grpc.insecure_channel("0.0.0.0:55555") as channel:
    VALStub(channel).GetValue(GetValueRequest())
```
Running this script should result in the panic on the 
```bash
# ...
2024-12-27T15:51:55.842868Z  INFO databroker::grpc::server: Authorization is not enabled.
thread 'tokio-runtime-worker' panicked at databroker/src/grpc/kuksa_val_v2/val.rs:856:37:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/databroker/src/grpc/kuksa_val_v2/val.rs
+++ b/databroker/src/grpc/kuksa_val_v2/val.rs
@@ -853,6 +853,10 @@ async fn get_signal(
     signal_id: Option<proto::SignalId>,
     broker: &AuthorizedAccess<'_, '_>,
 ) -> Result<i32, tonic::Status> {
+    if signal_id.is_none() {
+        return Err(tonic::Status::invalid_argument("No SignalId provided"));
+    }
+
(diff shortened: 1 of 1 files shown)


SebastianSchildt (approved):
Thank you for your (first (of many?)) contribution.

lgtm 🕹️

PR #126, opened 2025-01-28
A build workflow for the Release-With-Debug profile

mikehaller:
_Note: This PR is part of cleaning up #113 which got mixed up with devcontainer setup and more changes._

A build workflow for the `release-with-debug` profile, which creates a databroker binary with debug symbols enabled.
This is useful for the integration into an eBPF-based performance test environments.

As separate testers would need to rebuild databroker with that profile, which requires them to set up their own databroker development environment, we've decided for convenience to release debug versions of databroker for exactly that use case.

The additional `debugsymbols.txt` file in the archive contains the entry point for eBPF probes to instrument the function within databroker where the VSS Signals are being updated by providers:

```
DatabaseWriteAccess.update()
```


lukasmittag (changes requested):
looks good. Only the question about the updated `Cargo.lock` arises again :) Should we update this in this PR?

lukasmittag (approved):
LGTM

PR #127, opened 2025-01-28
Initial example for testing slow subscribers

mikehaller:
The example shows a subscriber which is intentionally slow in receiving updates.

When running in parallel with kuksa-perf, there are too many updates of the signal which the slow subscriber will not be able to keep up with and will start throwing away signal updates. When this happens, the channel on the databroker side will log a warning message and indicate how many signal updates were lost for the slow subscriber.


lukasmittag (changes requested):
LGTM. Only problem again the high level Cargo.lock that is updated the one under lib shall be kept.

lukasmittag (approved):
LGTM

PR #131, opened 2025-01-30
Disable keep-alive in tcp connection

lukasmittag:
We've seen 40 ms spikes periodically when meassuring latency. It seems if you disable keep-alive signals in tcp this issue goes away.

!!! DRAFT !!!

Not entirely sure this is the best way to go since channels might be closing the connection. See https://grpc.io/docs/guides/performance/#general 

Will do some further tests.


lukasmittag:
already done in #137

PR #135, opened 2025-02-06
Pre-Commit hook for cargo fmt and clippy

mikehaller:
To prevent checks on GitHub PullRequests to fail due to formatting and linting errors (which also may be introduced or changed by updated cargo versions), they are added to the pre-commit hook by this PR.

Did not add any other cargo build, compile or test steps, as these are much more expensive and would increase the wait time for commits unnecessarily.


lukasmittag (approved):
LGTM

PR #136, opened 2025-02-10
Fix tag names for debug build

SebastianSchildt:
Fixes debug builds, they were pointing to non existant tags/repositories for quay.io

See 

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5628


lukasmittag (approved):
LGTM

PR #137, opened 2025-02-11
Reduce latency and spikes by tunning TCP socket

rafaeling:


databroker/src/grpc/server.rs
```diff
@@ -183,8 +197,9 @@ where
     broker.start_housekeeping_task();
 
     let mut server = Server::builder()
-        .http2_keepalive_interval(Some(Duration::from_secs(10)))
-        .http2_keepalive_timeout(Some(Duration::from_secs(20)));
+        .http2_adaptive_window(Some(true))
+        .http2_keepalive_interval(None)
+        .http2_keepalive_timeout(None);
```
mikehaller, line 202: Really setting the timeout to None?

databroker/src/grpc/server.rs
```diff
@@ -112,7 +113,20 @@ where
...
+    let socket = Socket::new(Domain::IPV4, Type::STREAM, Some(Protocol::TCP))?;
+    
+    socket.set_linger(None)?;
+    socket.set_nonblocking(true)?;
+
+    socket.set_quickack(true)?;
+    socket.set_nodelay(true)?;
+    
+    socket.bind(&socket_addr.into())?;
+    socket.listen(128)?;
```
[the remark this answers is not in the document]
lukasmittag: https://docs.rs/socket2/latest/socket2/struct.Socket.html#examples see this. As of my knowledge and what I found out this is the backlog of listen https://man7.org/linux/man-pages/man2/listen.2.html for the corresponding os. Backlog means here how man connections can be queued/accepted by the socket See this post 128 seems to be the deafult linux value (https://www.linuxjournal.com/files/linuxjournal.com/linuxjournal/articles/023/2333/2333s2.html).

databroker/src/grpc/server.rs
lukasmittag: if we set a timeout without keep-alive pings this could lead to closing an `idle` connection if there is nothing sent or received for the timeout specified time. Say we set  `.http2_keepalive_timeout(20s);` and `.http2_keepalive_interval(None)` then after 20 s if there's no communication the connection will be closed. Ithink if we disable keep-alive we should set both to `None`. The `Danger` would be that we keep idle connections open for a very long time and they might be broken (e.g. we encounter broken connections only if we try to send something again). 

This could possibly lead to ressource leaks as the server may still hold resources (like memory, file descriptors, etc.) associated with the connection and the request. If the server is not aware that the client has disconnected, it may continue to allocate resources for that connection, leading to resource leaks.

Maybe setting the keepalive timeout to 1 min? or even keep-alive interval to 1 min and timeout to 2 could be a solution.

lukasmittag (approved):
LGTM

SebastianSchildt (approved):
lgtm 🚋

PR #140, opened 2025-02-25
Add kuksa.val.v2 Support for Kuksa Rust SDK

wba2hi:


lib/kuksa_val_v2/src/lib.rs
```diff
@@ -0,0 +1,1508 @@
...
+    ///                   e.g. if sending an unsupported enum value
+    ///              - if the published value is out of the min/max range specified
+    ///
+    ///    - Provider returns BatchActuateStreamResponse <- Databroker sends BatchActuateStreamRequest
+    ///        No error definition, a BatchActuateStreamResponse is expected from provider.
+    ///
+    pub async fn open_provider_stream(
+        &mut self,
+        receiver_stream: ReceiverStream<OpenProviderStreamRequest>,
+    ) -> Result<Streaming<OpenProviderStreamResponse>, ClientError> {
```
wba2hi, line 317: unsure if this is the "best" / "most convenient" signature.

Right now it must be used like this:
```
        let (sender, receiver) = tokio::sync::mpsc::channel(4);
        let receiver_stream = ReceiverStream::new(receiver);
        let mut stream = client.open_provider_stream(receiver_stream).await.unwrap();

        let provide_actuation_request = OpenProviderStreamRequest {
            action: Some(Action::ProvideActuationRequest(ProvideActuationRequest {
                actuator_identifiers: vec![
                    SignalId {
                        signal: Some(Path(actuator_2.to_string())),
                    },
                    SignalId {
                        signal: Some(Path(actuator_1.to_string())),
                    },
                ],
            })),
        };

        sender.send(provide_actuation_request).await.unwrap();
```

lukasmittag: one option would be to make a helper func that creates the stream and returns the sender obejct which you then could plug in to the calls and construct the request. 
Dummy code:
```
let sender, receiver = open_provider_stream( ....)
client.provide_actuation(sender, paths)
```
Another option would be that KuksaClientV2 can hold these sender and receiver as struct members and therefor you just need to call open_provider_stream() then the sender, receiver gets stored to handle stuff internally. E.g. just call provide_actuation() and inside the function it calls self.sender.send. Of course this would get difficult if you want to act on BatchActuateStreamRequests but for sending data to databroker it should work.

lukasmittag (approved):
As a first version of the rust sdk for kuksa.val.v2 nice! 

Points to discuss:
* how to handle bi-di streaming?
* could there be helper functions that you do not need to know the structure of `Value` or `Datapoint` 
* how could we leverage the performance upside with using ids instead of string paths?
* Is this the right place for these integration tests (if databroker behavior changes this will fail)

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -0,0 +1,1508 @@
...
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* SPDX-License-Identifier: Apache-2.0
+********************************************************************************/
+pub use databroker_proto::kuksa::val as proto;
+use databroker_proto::kuksa::val::v2::signal_id::Signal::Path;
+use databroker_proto::kuksa::val::v2::{
+    BatchActuateRequest, GetServerInfoRequest, GetValuesRequest, Metadata,
+    OpenProviderStreamResponse, SignalId, Value,
+};
```
lukasmittag, line 18: either remove use databroker_proto_kuksa::val as proto 5 lines above. or put this into use proto:v2: in line 25

lukasmittag (changes requested):

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -0,0 +1,1508 @@
...
+            signal_paths: paths,
+            buffer_size: buffer_size.unwrap_or(0),
+        };
+
+        match client.subscribe(subscribe_request).await {
+            Ok(response) => Ok(response.into_inner()),
+            Err(err) => Err(Status(err)),
+        }
+    }
+
```
rafaeling, line 282: subscribe_by_id is missing here?

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -0,0 +1,1508 @@
...
+* SPDX-License-Identifier: Apache-2.0
+********************************************************************************/
+pub use databroker_proto::kuksa::val as proto;
+use databroker_proto::kuksa::val::v2::signal_id::Signal::Path;
+use databroker_proto::kuksa::val::v2::{
+    BatchActuateRequest, GetServerInfoRequest, GetValuesRequest, Metadata,
+    OpenProviderStreamResponse, SignalId, Value,
+};
+use databroker_proto::kuksa::val::v2::{Datapoint, OpenProviderStreamRequest};
+use http::Uri;
```
rafaeling, line 20: Can you group proto references together and others like http with tokio, tonic, etc

databroker/Cargo.toml
```diff
@@ -104,3 +104,6 @@ tonic-mock = "0.3.0"
 [[test]]
 name = "read_write_values"
 harness = false
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
```
rafaeling, line 109: Why did you included this here and was not there before?

rafaeling (commented):
Looks good to me, really extensive unit tests covering, just subscribe_by_id missing in skd

databroker/Cargo.toml
wba2hi: without this statement `cargo clippy` is failing for me with the following warning: 

```
warning: unexpected `cfg` condition name: `test`
   --> databroker/src/query/executor.rs:433:7
    |
433 | #[cfg(test)]
    |       ^^^^
    |
    = help: consider using a Cargo feature instead
    = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
             [lints.rust]
             unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
    = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(test)");` to the top of the `build.rs`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
```

Don't know why this is a only-me problem? Maybe due to the cargo version?

```
kuksa-databroker git:(feature/rust-sdk-kuksa-val-v2-support) ✗ cargo --versio
n
cargo 1.84.1
```

lib/databroker-examples/examples/kuksa_val_v2.rs
```diff
@@ -0,0 +1,264 @@
...
+                Err(err) => {
+                    println!("Error subscribing to ids: {:?}", err);
+                }
+            }
+        }
+        Err(err) => {
+            println!("Error: Could not resolve ids for path: {:?}", err);
+        }
+    }
+}
```
wba2hi, line 233: This method still feels too complicated,...?

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+
+        for path in vss_paths {
+            let vec = self.list_metadata(path, "*").await?;
+            let metadata = vec.first().unwrap();
+
+            hash_map.insert(metadata.path.clone(), metadata.id);
+        }
+
+        Ok(hash_map)
+    }
```
wba2hi, line 471: Open for discussion:

Should we expose this or not? I chose a HashMap here, to make it easier when used together with subscribe_by_id: https://github.com/eclipse-kuksa/kuksa-databroker/pull/140/files#diff-d203cebe4619670778c81e9cec39a975575be5ef4b305f7b20531fdbe8c35931R201

lib/databroker-examples/examples/kuksa_val_v2.rs
```diff
@@ -0,0 +1,264 @@
...
+                            };
+                            if let Some(average_speed) = response.entries.get(path_avg_speed) {
+                                // do something with average_speed
+                                println!("{}: {:?}", path_avg_speed, average_speed);
+                            };
+                        }
+                        None => {
+                            // no subscribe response
+                            println!("No SubscribeResponse for {}", path_speed);
+                        }
```
wba2hi, line 169: Can this even occur?

wba2hi:
> As a first version of the rust sdk for kuksa.val.v2 nice!
> 
> Points to discuss:
> 
>     * how to handle bi-di streaming?

I tried to improve this stuff a bit, now we create the receiver stream inside the method itself and then return the corresponding sender and receiver using an OpenProviderStream object. this should simplify it's usage and still allow enough flexibility. 

 
>     * could there be helper functions that you do not need to know the structure of `Value` or `Datapoint`

Didn't scratch that topic yet, left for follow up ticket. Only way to solve this is probably to create a wrapper class, which exposes similar functionality as Datapoint / Value.
 
>     * how could we leverage the performance upside with using ids instead of string paths?

I fiddled around but this feels like it might take a bit longer to implement. I tried adding an IdCache which will be purged on disconnect, but had mostly issue with rust specific stuff again.   

>     * Is this the right place for these integration tests (if databroker behavior changes this will fail)

I see your point, probably the tests would be better placed somewhere else (and then maybe not test the SDK functionality but the proto methods) however, if databroker behavior changes which will break the test it will directly show up in Github Actions workflows and can / must be fixed by whoever introduced the breaking changes.

mikehaller:
Nice to see the Kuksa Rust SDK growing, thank you

lib/databroker-examples/examples/kuksa_val_v2.rs
rafaeling: You could reduce maybe some print functions and use instead `is_some` function to avoid the unwrapping.
However Results and match clauses look so long like this.

lib/kuksa_val_v2/src/lib.rs
lukasmittag: hm I would be for exposing this maybe helps for something else as well :)

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+
+    fn convert_to_vector(values: HashMap<String, Value>) -> Vec<ActuateRequest> {
```
lukasmittag, line 472: this name confuses me it is not speaking to me :) can we rename to something like create_actuate_request() or value_to_actuate_request() ?

lib/databroker-examples/examples/kuksa_val_v2.rs
wba2hi: Is this path necessary when dealing with open_provider_stream, like here: https://github.com/eclipse-kuksa/kuksa-databroker/pull/140/files#diff-d203cebe4619670778c81e9cec39a975575be5ef4b305f7b20531fdbe8c35931R136 ? 
From my understanding this is something like a callback, if there is no message, then stream.message() should not continue with an empty option. However AFAIK we are not sending empty responses back in open_provider_stream, do you think removing the branch might be conflicting or safe to remove?

lib/kuksa_val_v2/src/lib.rs
wba2hi: Indeed, that method was much to generic. Renamed it to "convert_to_actuate_requests", this should be more fitting.

lukasmittag (approved):
LGTM

PR #141, opened 2025-02-26
Extend bidirectional provider by introducing sample interval filtering

rafaeling:


proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,24 @@ message Value {
...
+
+enum FilterError {
+  FILTER_ERROR_CODE_UNSPECIFIED   = 0;
+  FILTER_ERROR_CODE_OK            = 1;
+  FILTER_ERROR_CODE_NETWORK_ERROR = 2;
+  FILTER_ERROR_CODE_OVERLOAD      = 3;
+}
+
+message Filter {
+  // Duration of the active call. If it is not set, call will least for ever.
```
lukasmittag, line 62:
```suggestion
  // Duration of the active call. If it is not set, call will last forever.
```

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,24 @@ message Value {
...
+  int32 duration_ms                  = 1;
+  // Max desired sample update interval.
+  SampleInterval max_sample_interval = 2;
```
lukasmittag, line 65: I would use int32 inline

proto/kuksa/val/v2/types.proto
```diff
@@ -74,6 +92,13 @@ enum ErrorCode {
   ERROR_CODE_PERMISSION_DENIED = 4;
 }
 
+enum CommandError {
```
lukasmittag, line 95: No longer needed?

proto/kuksa/val/v2/val.proto
```diff
@@ -268,16 +275,13 @@ message PublishValueRequest {
   Datapoint data_point = 2;
 }
 
-message PublishValueResponse {
-}
-
 message PublishValuesRequest {
-  int32 request_id                 = 1; /// Unique request id for the stream that can be used to identify the response.
+  uint64 request_it                 = 1; /// Unique request id for the stream that can be used to identify the command.
```
lukasmittag, line 279:
```suggestion
  uint64 request_id                 = 1; /// Unique request id for the stream that can be used to identify the command.
```

lukasmittag (changes requested):

BjoernAtBosch (changes requested):

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,24 @@ message Value {
...
+enum FilterError {
+  FILTER_ERROR_CODE_UNSPECIFIED   = 0;
+  FILTER_ERROR_CODE_OK            = 1;
+  FILTER_ERROR_CODE_NETWORK_ERROR = 2;
+  FILTER_ERROR_CODE_OVERLOAD      = 3;
+}
+
+message Filter {
+  // Duration of the active call. If it is not set, call will least for ever.
+  int32 duration_ms                  = 1;
```
BjoernAtBosch, line 63: Better use uint - what would the meaning of negative values be?
```suggestion
  uint32 duration_ms                 = 1;
```

proto/kuksa/val/v2/types.proto
BjoernAtBosch, line 65: It's minimum (accepted) interval, i.e. maximum accepted rate
```suggestion
  SampleInterval min_sample_interval = 2;
```

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,24 @@ message Value {
   }
 }
 
+message SampleInterval {
+  int32 max_interval_ms = 1;
```
BjoernAtBosch, line 51: If we keep this as separate structure, I'd remove the max to make this usable to possibly add a maximum interval (.i.e. minimum rate) in the `Filter` message ...
```suggestion
  int32 interval_ms = 1;
```

BjoernAtBosch (changes requested):

wba2hi (changes requested):

proto/kuksa/val/v2/types.proto
rafaeling: Since it will be also used in Metadata I decided to create its own message to keep the same semantic. What do you think?

proto/kuksa/val/v2/types.proto
rafaeling: Nop

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,24 @@ message Value {
...
+}
+
+enum FilterError {
```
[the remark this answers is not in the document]
rafaeling: The purpose of this error enum was to transmit Provider's context errors, like can network error, provider overloaded, etc

proto/kuksa/val/v2/val.proto
```diff
@@ -44,6 +45,7 @@ service VAL {
   //   PERMISSION_DENIED if access is denied for any of the requested signals.
   //   INVALID_ARGUMENT if the request is empty or provided path is too long
   //       - MAX_REQUEST_PATH_LENGTH: usize = 1000;
+  //   UNAVAILABLE if there is no provider currently providing the signals
   //
```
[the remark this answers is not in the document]
rafaeling: Decided after meeting to remove it and keep the non availability of provider as None in datapoint

proto/kuksa/val/v2/types.proto
```diff
@@ -74,6 +92,13 @@ enum ErrorCode {
   ERROR_CODE_PERMISSION_DENIED = 4;
 }
 
+enum CommandError {
+  COMMAND_ERROR_CODE_UNSPECIFIED   = 0; // Default value, never to be explicitly set,
+  COMMAND_ERROR_CODE_OK            = 1;
+  COMMAND_ERROR_CODE_NETWORK_ERROR = 2;
+  COMMAND_ERROR_CODE_OVERLOAD      = 3;
+}
+
```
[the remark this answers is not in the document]
rafaeling: removed

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,24 @@ message Value {
...
+message SampleInterval {
+  int32 max_interval_ms = 1;
+}
+
+enum FilterError {
+  FILTER_ERROR_CODE_UNSPECIFIED   = 0;
+  FILTER_ERROR_CODE_OK            = 1;
+  FILTER_ERROR_CODE_NETWORK_ERROR = 2;
+  FILTER_ERROR_CODE_OVERLOAD      = 3;
+}
```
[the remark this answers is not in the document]
rafaeling: WIll added it

proto/kuksa/val/v2/val.proto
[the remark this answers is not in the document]
rafaeling: yes, thanks

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,24 @@ message Value {
   }
 }
 
+message SampleInterval {
```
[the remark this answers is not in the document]
rafaeling: It is used in Metadata and Filter messages, wanted to keep same semantic and meaning between both messages.

proto/kuksa/val/v2/val.proto
```diff
@@ -268,16 +271,13 @@ message PublishValueRequest {
...
-
 message PublishValuesRequest {
-  int32 request_id                 = 1; /// Unique request id for the stream that can be used to identify the response.
+  uint64 request_id                 = 1; /// Unique request id for the stream that can be used to match the corresponding response.
   map<int32, Datapoint> data_points = 2;
 }
 
 message PublishValuesResponse {
-  int32 request_id         = 1;
+  uint64 request_id        = 1; /// Unique request id for the stream that can be used to match the corresponding request.
```
BjoernAtBosch, line 280: Missed that yesterday: Do we really need a 64 bit request id? This would mean, we expect that request out of 4 billion could potentially overlap ...
BTW: Wouldn't that be an incompatible interface change - requiring to proceed to v3?

BjoernAtBosch (changes requested):

proto/kuksa/val/v2/val.proto
```diff
@@ -131,6 +131,7 @@ service VAL {
   //   NOT_FOUND if the specified root branch does not exist.
   //   UNAUTHENTICATED if no credentials provided or credentials has expired
   //   INVALID_ARGUMENT if the provided path or wildcard is wrong.
+  //   UNAVAILABLE if there is no provider currently providing the signal -> do we want this error for list metadata?
   //
   rpc ListMetadata(ListMetadataRequest) returns (ListMetadataResponse);
```
BjoernAtBosch, line 137: Should we add an error to `PublishValue` for the case that the addressed signal is already "claimed" by another provider (via OpenProviderStream -> ProvideSignalRequest). E.g.
`  //   RESOURCE_EXHAUSTED if the addressed signal is already occupied by another provider`
Or what is the current behavior of the broker in those cases?

proto/kuksa/val/v2/val.proto
```diff
@@ -288,6 +288,13 @@ message ProvideActuationRequest {
 message ProvideActuationResponse {
 }
 
+message ProvideSignalRequest {
+  map<int32, SampleInterval> signals_sample_intervals = 1;
+}
+
+message ProvideSignalResponse {
+}
+
```
BjoernAtBosch, line 297: Sorry, I didn't recognize that the ProvideSignalRequest/Response is also new. We should specify the behavior when a provider freshly claims a signal: Does he immediately start to send updates or does he wait until a UpdateFilterRequest comes in? What about "legacy" providers not knowing the filter concept? Or should we send back a filter map in the response (like in the UpdateFilterRequest) which is defining the initial filter?

And BTW: We did not specify the behavior if already another provider has claimed signals for actuation or provisioning. Maybe there should be some respective "error signalling" in the responses.

proto/kuksa/val/v2/val.proto
```diff
@@ -298,6 +309,28 @@ message BatchActuateStreamResponse {
   Error error = 2;
 }
 
+message UpdateFilterRequest {
+  // Databroker sends filters to provider.
```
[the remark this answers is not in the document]
rafaeling: Each provider opens its OpenBidirectional call, so request and response a kind of "binded".
But maybe will benefit to have it (for the wait response before sending case), will add it.

proto/kuksa/val/v2/val.proto
```diff
@@ -131,6 +131,7 @@ service VAL {
   //   NOT_FOUND if the specified root branch does not exist.
   //   UNAUTHENTICATED if no credentials provided or credentials has expired
   //   INVALID_ARGUMENT if the provided path or wildcard is wrong.
+  //   UNAVAILABLE if there is no provider currently providing the signal -> do we want this error for list metadata?
```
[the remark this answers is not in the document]
rafaeling: You are right, thanks

proto/kuksa/val/v2/val.proto
```diff
@@ -268,16 +271,13 @@ message PublishValueRequest {
   Datapoint data_point = 2;
 }
 
-message PublishValueResponse {
-}
-
```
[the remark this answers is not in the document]
rafaeling: mein Fehler

proto/kuksa/val/v2/val.proto
rafaeling: I will make it uint32, to avoid negative request ids.

proto/kuksa/val/v2/val.proto
rafaeling: Actually was no the same as for actuators claiming? first who claim actuator/sensor wins?
Do we really want to add an error to the response?

proto/kuksa/val/v2/val.proto
rafaeling: The behavior is that, as long as there isnt any signal consumer, no datapoints will be transmitted from provider to databroker.
Provider would claim signal, but it will start publishing to databroker upon FilterRequest received.

You are right, we did not consider yet that use case, but I think it was out of scope for now more than one provider.

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,40 @@ message Value {
...
+
+enum FilterError {
+  FILTER_ERROR_CODE_UNSPECIFIED      = 0;
+  FILTER_ERROR_CODE_NETWORK_ERROR    = 1;
+  FILTER_ERROR_CODE_OVERLOAD         = 2;
+  FILTER_ERROR_CODE_UNKNOW_SINGAL_ID = 3;
+}
+
+message Filter {
+  // Duration of the active call. If it is not set, call will last for ever.
```
datng23, line 76: Hi @rafaeling,
Just curious question.
I got the concept of Filter to mange network load based on application expectation rate. For my knowledge there would be some policy we can apply:
1) Limiting the rate.
2) On change (w/wo threshold).
3) On change else rate (as heartbeat).

Is that same purpose for SampleInterval and OperationFilter?

rafaeling: Hi Dat,

yes, the idea of sample interval is limiting the rate.
We have just decided to postpone for now the OperationFilter for points 2 and 3, it needs further discussions

proto/kuksa/val/v2/val.proto
```diff
@@ -306,24 +316,31 @@ message BatchActuateStreamResponse {
...
 }
 
 // Only returned if there is a filter error on provider side
 message UpdateFilterResponse {
-  FilterError filter_error = 1;
+  uint32 request_id         = 1; /// Unique request id for the stream that can be used to match the corresponding request.
+  FilterError filter_error  = 2;
+}
+
+// Only returned if there is a error on provider side
```
BjoernAtBosch, line 332:
```suggestion
// Send to indicate an error on provider side
```

proto/kuksa/val/v2/val.proto
```diff
@@ -266,18 +272,22 @@ message ListMetadataResponse {
...
 }
 
+message PublishValueResponse {
+
+}
+
 message PublishValueRequest {
   SignalID signal_id   = 1;
   Datapoint data_point = 2;
 }
```
BjoernAtBosch, line 283: Just changing the order (definition of request before response)
```suggestion
message PublishValueRequest {
  SignalID signal_id   = 1;
  Datapoint data_point = 2;
}

message PublishValueResponse {
}

```

BjoernAtBosch (commented):
Just two minors

proto/kuksa/val/v2/types.proto
```diff
@@ -47,6 +47,29 @@ message Value {
   }
 }
 
+message SampleInterval {
+  uint32 interval_ms = 1;
+}
+
+enum FilterError {
+  FILTER_ERROR_CODE_UNSPECIFIED      = 0;
+  FILTER_ERROR_CODE_UNKNOW_SINGAL_ID = 1;
```
BjoernAtBosch, line 56:
```suggestion
  FILTER_ERROR_CODE_UNKNOWN_SINGAL_ID = 1;
```

proto/kuksa/val/v2/types.proto
```diff
@@ -113,7 +136,9 @@ message Metadata {
   Value allowed_values                 = 17; // Must be of array type
   Value min                            = 18;
   Value max                            = 19;
-
+  
+  // Minimum sample interval at which its provider will publish the signal value
```
BjoernAtBosch, line 140:
```suggestion
  // Minimum sample interval at which its provider can publish the signal value
```

proto/kuksa/val/v2/val.proto
```diff
@@ -318,6 +369,14 @@ message OpenProviderStreamResponse {
...
     // Send a batch actuate request to a provider.
     BatchActuateStreamRequest batch_actuate_stream_request = 3;
+    // Response to a provide sensor request.
+    ProvideSignalResponse provide_signal_response          = 4;
+    // Filter request
+    UpdateFilterRequest update_filter_request              = 5;
+    // GetValue request from client forwarded to provider
+    GetProviderValueRequest get_provider_value_request     = 6;
+    // Indication of error on provider side
+    ProviderErrorIndication provider_error_indication      = 7;
```
BjoernAtBosch, line 379:
```suggestion
```

proto/kuksa/val/v2/val.proto
```diff
@@ -307,6 +352,12 @@ message OpenProviderStreamRequest {
     // Sent to acknowledge the acceptance of a batch actuate
     // request.
     BatchActuateStreamResponse batch_actuate_stream_response = 3;
+    // Inform server of a signal this provider provides.
+    ProvideSignalRequest provide_signal_request              = 4;
+    // Update filter response
+    UpdateFilterResponse update_filter_response              = 5;
+    // GetValue response
+    GetProviderValueResponse get_provider_value_response     = 6;
   }
```
BjoernAtBosch, line 361: This goes from provider to broker, isn't it?
```suggestion
    // Indication of error on provider side
    ProviderErrorIndication provider_error_indication        = 7;
  }
```

BjoernAtBosch (changes requested):
One can always find somethig ;-)

BjoernAtBosch (commented):
API LGTM

wba2hi (dismissed):
LGTM

BjoernAtBosch (approved):
LGTM

wba2hi (approved):

databroker/src/grpc/kuksa_val_v2/val.rs
```diff
@@ -693,6 +694,18 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                                todo!();
+                                            }
+                                            Some(UpdateFilterResponse(_update_filter_response)) => {
+                                                todo!();
+                                            }
+                                            Some(GetProviderValueResponse(_get_provider_value_response)) => {
+                                                todo!();
+                                            }
+                                            Some(ProviderErrorIndication(_provide_error_indication)) => {
+                                                todo!();
```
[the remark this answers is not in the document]
rafaeling: The implementation of this code is here: https://github.com/boschglobal/kuksa-databroker/tree/bidirectional-provider-implementation

This PR just cares about protobuf files. I will create a new PR with the implementation once this gets merged.

doc/user_guide.md
```diff
@@ -72,7 +72,6 @@ docker run --rm -it -p 55555:55555 ghcr.io/eclipse-kuksa/kuksa-databroker:main -
 > docker run --rm -it -p 55556:55555 ghcr.io/eclipse-kuksa/kuksa-databroker:main --insecure
 > ```
 >
-> Please refer to [this support forum post](https://developer.apple.com/forums/thread/671197) for additional information.
```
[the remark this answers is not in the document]
rafaeling: Link died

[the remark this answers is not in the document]
lukasmittag: I see

lukasmittag (approved):
LGTM

PR #142, opened 2025-02-26
Create a ClientTrait for RustSDK

lukasmittag:
For handling the Rust SDK and preparing to make it an own crate this introduces a `ClientTrait`. With this the maintenance of the soon-to-be Rust SDK will be easy and manageable.


databroker-cli/src/kuksa_cli.rs
```diff
@@ -432,24 +423,12 @@ pub async fn kuksa_main(_cli: Cli) -> Result<(), Box<dyn std::error::Error>> {
...
-                                                "metadata",
-                                                format_args!("Error {msg:?}"),
-                                            )?;
-                                        }
+                                    if let Some(entries) =
+                                        handle_get_metadata(vec![], &mut client).await.unwrap()
+                                    {
+                                        interface.set_completer(Arc::new(
+                                            CliCompleter::from_metadata(&entries),
+                                        ));
```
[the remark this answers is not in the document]
lukasmittag: yes

lib/common/src/lib.rs
```diff
@@ -41,6 +41,95 @@ pub enum ClientError {
...
+    type SubscribeType;
+    type PublishResponseType;
+    type GetResponseType;
+    type SubscribeResponseType;
+    type ProvideResponseType;
+    type ActuateResponseType;
+    type MetadataResponseType;
+
+    // from deeply embedded layer providing sensor values (to keep backwards compatibility the naming is different for the corresponding interfaces)
+    // if we do not want to put in the effort just give an unimplemented error for the function
```
rafaeling, line 58: Really like the idea of having traits here, just one thing:
what about splitting interfaces? like creating a generic interface which cover common cases from all the APIs like subscribe, publish, etc and an specific traits which covers the target_values use cases, etc?
Something like:
`trait GenericSDK {
    fn subscribe(&self);
}`
`trait TargetValueSDK: GenericSDK {
    fn set_target_values(&self);
}`

If I am a sdk user for kuksa_val_v2, it would be confuse why to see set_target_values, etc methods

lukasmittag: hmm yes but I thought a little bit about managing backwards compatibility. For the sdk agree that it looks strange for people. And for now we might not need backwards compatibility. However I will think about generic sdk and additions.

lib/common/src/conversion.rs
```diff
@@ -26,6 +27,29 @@ use std::collections::HashMap;
 
 // Idea: in the future we could use databroker internal datapoint structure and define it here then the conversion from databroker can be reused
 
+fn find_common_root(paths: Vec<String>) -> String {
```
wba2hi, line 30: This method is perfectly fit for a unit test :)

lib/common/src/conversion.rs
```diff
@@ -486,225 +510,18 @@ impl ConvertToV1<Option<protoV1::datapoint::Value>> for SDVprotoV1::Datapoint {
...
-
-impl ConvertToV2<ActuateResponseTypeV2> for ActuateResponseTypeV1 {
-    fn convert_to_v2(self) -> ActuateResponseTypeV2 {
-        self
-    }
-}
-
 impl ConvertToV1<MetadataResponseTypeV1> for MetadataResponseTypeV2 {
     fn convert_to_v1(self) -> MetadataResponseTypeV1 {
         //log::warn("This method is deprecated and conversion can not figure out which path the datapoint maps to. Dev must do this in his own code. E.g. by going through the requested paths.");
```
wba2hi, line 527: Needed? not needed?

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -566,9 +553,13 @@ impl ClientTraitV2 for KuksaClientV2{
                 };
                 Ok(server_info)
             }
-            Err(err) => Err(Status(err)),
+            Err(err) => Err(ClientError::Status(err)),
         }
     }
+
+    async fn provide_actuation(&mut self, _path: Self::PathType) -> Result<Self::ProvideResponseType, ClientError> {
+        unimplemented!()
```
wba2hi, line 561: Why unimplemented? shouldn't provide_actuation be supported?

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -54,12 +54,12 @@ impl KuksaClientV2 {
     ///
     pub async fn resolve_ids_for_paths(
         &mut self,
-        vss_paths: Vec<&str>,
+        vss_paths: Vec<String>,
```
wba2hi, line 57: Could you elaborate on this change? Is it better to use a "String" over a "&str"?

wba2hi (changes requested):
No functional test has been done yet

lib/common/src/conversion.rs
```diff
@@ -0,0 +1,1004 @@
...
+    }
+}
+
+impl ConvertToV1<MetadataResponseTypeV1> for MetadataResponseTypeV2 {
+    fn convert_to_v1(self) -> MetadataResponseTypeV1 {
+        //log::warn("This method is deprecated and conversion can not figure out which path the datapoint maps to. Dev must do this in his own code. E.g. by going through the requested paths.");
+        let transformed_vec = self
+            .iter()
+            .map(|metadata| protoV1::DataEntry {
+                path: "Unknwon".to_string(),
```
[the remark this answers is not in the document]
lukasmittag: becasue metadata of kuksa.val.v1 does not have the path in the metadata :( did  a comment some lines above if I remember correctly.

lib/common/src/conversion.rs
```diff
@@ -0,0 +1,1004 @@
...
+//     }
+// }
+
+// fn convert_value_restriction(
+//     restriction: Option<protoV1::ValueRestriction>,
+// ) -> Option<SDVprotoV1::Allowed> {
+//     restriction.map(|restriction| SDVprotoV1::Allowed {
+//         values: Some(convert_to_allowed(restriction)),
+//     })
+// }
```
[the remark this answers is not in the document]
lukasmittag: If you want to write a conversion from SDV to kuksa.val.v1 you would need this but since we not intend to have "forward" compatibility not needed will remove.

lib/common/src/types.rs
```diff
@@ -0,0 +1,77 @@
...
+pub type GetResponseTypeV2 = protoV2::Datapoint;
+pub type MultipleGetResponseTypeV2 = Vec<GetResponseTypeV2>;
+pub type SubscribeResponseTypeV2 = tonic::Streaming<protoV2::SubscribeResponse>;
+pub type SubscribeByIdResponseTypeV2 = tonic::Streaming<protoV2::SubscribeByIdResponse>;
+pub type ProvideResponseTypeV2 = ();
+pub type ActuateResponseTypeV2 = ();
+pub type OpenProviderStreamResponseTypeV2 = OpenProviderStream;
+pub type MetadataTypeV2 = (PathTypeV2, PathTypeV2);
+pub type MetadataResponseTypeV2 = Vec<protoV2::Metadata>;
+pub type ServerInfoTypeV2 = ServerInfo;
```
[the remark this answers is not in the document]
lukasmittag: hmm, think of it more to handle the conversions and now what you want to do there. i know it feels arbitrary. One could argue you can just use the same naming as in the API's. Thats what I thought about but was not sure this does not suggest then something different. So for example `publish_value` have a `PublishValueRequestType` and `PublishValueResponseType` same for the other functions - `XRequestType` `XResponseType`. Would that make more sense to you? I introduced `PathType` because I felt like every underlying function should handle the input path as the same in ONE ClientTrait however it could differ BETWEEN Traits.

lib/common/src/conversion.rs
lukasmittag: Yes, will add some unit test in the next iteration :D

lib/kuksa_val_v2/src/lib.rs
lukasmittag: That's what I'm currently working on. The problem with references in rust and traits seems to be that lifetime handling is difficult. So I'm currently trying to solve this by using clones and hence no references. But I'm experimenting wit lifetimes now.

lib/common/src/conversion.rs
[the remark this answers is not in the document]
wba2hi: grafik
You sure? It seems to work oO

lib/kuksa_val_v2/src/lib.rs
lukasmittag: I experimented a little bit with references and lifetimes. Could not bring it in a solid state yet https://github.com/boschglobal/kuksa-databroker/tree/experiment/client_trait_refs_lifetimes so will continue fixing review findings due to the lack of time.

lib/kuksa_val_v2/src/lib.rs
lukasmittag: yes, but its one of these where you need an open_provider_stream element and we do not have the solution to that yet. Changed it to todo! so the macro will display that it is "not yet implemented" inested of "not implemented".

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -138,33 +141,49 @@ impl kuksa_common::ClientTraitV1 for KuksaClientV2 {
         &mut self,
         paths: Self::SubscribeType,
     ) -> Result<Self::SubscribeResponseType, ClientError> {
-        Ok(ClientTraitV2::subscribe(self, paths.convert_to_v2(), None).await.unwrap().convert_to_v1())
+        Ok(ClientTraitV2::subscribe(self, paths.convert_to_v2(), None)
+            .await
+            .unwrap()
+            .convert_to_v1())
```
wba2hi, line 147: will panic

lib/kuksa_val_v2/src/lib.rs
```diff
@@ -138,33 +141,49 @@ impl kuksa_common::ClientTraitV1 for KuksaClientV2 {
...
 
     async fn subscribe(
         &mut self,
         paths: Self::SubscribeType,
     ) -> Result<Self::SubscribeResponseType, ClientError> {
-        Ok(ClientTraitV2::subscribe(self, paths.convert_to_v2(), None).await.unwrap().convert_to_v1())
+        Ok(ClientTraitV2::subscribe(self, paths.convert_to_v2(), None)
+            .await
+            .unwrap()
+            .convert_to_v1())
```
wba2hi, line 157: will panic

lib/kuksa/src/lib.rs
```diff
@@ -104,83 +106,93 @@ impl KuksaClient {
...
-                .await
-            {
-                Ok(mut entry) => metadata_result.append(&mut entry),
-                Err(err) => return Err(err),
-            }
-        }
-
-        Ok(metadata_result)
+#[async_trait]
+impl kuksa_common::SDVClientTraitV1 for KuksaClient {
```
wba2hi, line 112: Asking myself here, should the KuksaClient (val.v1) really implement the SDVClientTrait? If there are similar methods I would probably re-declare them in the corresponding interface instead of "implementing all SDV stuff" where we potentially inherit a lot of unimplemented / unsupported / incompatible stuff. 

The more unimplemented / incompatible stuff is exposed on a general level, the less simple and useful is the abstraction.

lib/common/src/types.rs
```diff
@@ -0,0 +1,89 @@
...
+pub type SubscribeByIdTypeV2 = IdsTypeV2;
+pub type PublishResponseTypeV2 = ();
+pub type GetResponseTypeV2 = Option<protoV2::Datapoint>;
+pub type MultipleGetResponseTypeV2 = Vec<protoV2::Datapoint>;
+pub type SubscribeResponseTypeV2 = tonic::Streaming<protoV2::SubscribeResponse>;
+pub type SubscribeByIdResponseTypeV2 = tonic::Streaming<protoV2::SubscribeByIdResponse>;
+pub type ProvideResponseTypeV2 = ();
+pub type ActuateResponseTypeV2 = ();
+pub type OpenProviderStreamResponseTypeV2 = OpenProviderStream;
+pub type MetadataTypeV2 = (PathTypeV2, PathTypeV2);
```
wba2hi, line 63: this is vssPath: String, filter: String, right?
While it's true that PathTypeV2 is defined as a String I still would prefer to not simply re-use "PathTypeV2", but (PathTypeV2, String)

lib/common/src/conversion.rs
```diff
@@ -0,0 +1,1044 @@
...
+
+impl ConvertToV1<ActuateResponseTypeV1> for ActuateResponseTypeV2 {
+    fn convert_to_v1(self) -> ActuateResponseTypeV1 {
+        self
+    }
+}
+
+impl ConvertToV1<SubscribeResponseTypeV1> for SubscribeResponseTypeV2 {
+    fn convert_to_v1(self) -> SubscribeResponseTypeV1 {
+        unimplemented!("Not possible to convert stream objects!")
```
wba2hi, line 521: well, it actually is possible isn't it? read the messages from stream, convert them to a v1 object and then emitting them to a new stream. compare to databroker/kuksa_val_v1/val.rs -> convert_to_proto_stream but yeah I admit, that whatever relates to streaming stuff feels really overcomplicated in Rust.

wba2hi (changes requested):

lib/kuksa/src/lib.rs
lukasmittag: I agree, that was just the first attempt to keep backwards compatibility. If the interface is deprecated this would be changed to a deprecation message.

wba2hi (changes requested):
Functional Test: 
- could not test kuksa.val.v2, because it's not yet supported by databroker-cli
- kuksa.val.v1 worked
- sdv.databroker.v1 returned unimplemented for everything, but this issue already seems to persist on main?

wba2hi (approved):
LGTM

PR #149, opened 2025-03-26
Bidirectional provider implementation

rafaeling:
Introduced the FilterManager structure to manage filter subscriptions based on the lowest time interval.

Added new fields to the Subscriptions struct to manage ChangeSubscription and Providers by their UUID.

Modified sender: broadcast::Sender<EntryUpdates> in ChangeSubscription to sender: broadcast::Sender<Option<EntryUpdates>> to allow receiving None when the sender is gone. This enables returning an error (VISS case) or None to the consumer.

Implemented logic to propagate time-based filters to providers, notify subscriptions according to the time interval, and claim signals by providers during startup.

Added VISS subscription integration tests.


databroker/src/broker.rs
```diff
@@ -124,7 +136,8 @@ pub struct Database {
 pub struct Subscriptions {
     actuation_subscriptions: Vec<ActuationSubscription>,
     query_subscriptions: Vec<QuerySubscription>,
-    change_subscriptions: Vec<ChangeSubscription>,
+    change_subscriptions: HashMap<Uuid, ChangeSubscription>,
```
lukasmittag, line 139: why store uuid two times and not make another field id instead?

databroker/src/broker.rs
```diff
@@ -202,8 +239,10 @@ pub struct QuerySubscription {
 
 pub struct ChangeSubscription {
     entries: HashMap<i32, HashSet<Field>>,
-    sender: broadcast::Sender<EntryUpdates>,
+    sender: broadcast::Sender<Option<EntryUpdates>>,
```
lukasmittag, line 242: this is a rather big concept change right?

databroker/src/filter/filter_manager.rs
```diff
@@ -0,0 +1,256 @@
...
+            })
+            .collect()
+    }
+
+    ///
+    /// Add a new filter to each signal_id in the database and returns
+    /// the new update filter map containing ONLY signals_ids along with
+    /// their new lowest interval_ms.
+    ///
+    pub fn add_new_update_filter(
```
lukasmittag, line 50: align naming e.g. `add_new_update_filter` and `remove_filter_by_subscription_uuid` make it clear that this is the same filter we're talking about

integration_test/gen_proto/kuksa/val/v1/types_pb2.py
lukasmittag, line 1: please do not check in generated files

integration_test/viss/pytest.ini
```diff
@@ -17,5 +17,6 @@ bdd_features_base_dir = "features"
 viss_ws_base_url = ws://localhost:8090
 viss_http_base_url = http://localhost:8090
 viss_mqtt_base_url = mqtt://localhost:1883
+viss_grpc_base_url = localhost:55555
```
lukasmittag, line 20: grpc:// would be better

lukasmittag (changes requested):

integration_test/gen_proto/kuksa/val/v1/types_pb2.py
rafaeling: For sdv API were already there for the integration test, if I dont add them integration test would not work

terbshaeusser-itk:
I also noticed during some testing that no UpdateFilter request is sent to the provider if no timebased filter is given in the VISS subscription message. Is this intentional? How would the provider in this case know, when to send new values?

databroker/src/viss/v2/types.rs
```diff
@@ -212,6 +214,12 @@ pub struct StaticMetadataFilter {
     // pub parameters: Option<StaticMetadataParameters>,
 }
 
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TimebasedFilter {
+    pub period: u32,
```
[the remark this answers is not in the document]
rafaeling: Fixed, thanks

rafaeling:
> > You can understand it in two ways:
> > ```
> > * repeatedly return subscription event messages -> if no timebased filter is set sent continously at the rate you want
> > * as specified by the [filter request](https://www.w3.org/TR/viss2-core/#filter-request) described in the [[viss2-core] -> if timebased filter is set accordingly to it
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but since it is in the same sentence it leaves doubts if timebased filter is mandatory.
> 
> I think there is a misunderstanding here. Our goal was not only to allow the client to basically throttle the sending of values but also to allow subscription based systems as a provider. It should be possible to have a provider that only sends values as long as at least one client is subscribed. This would help to reduce system load and unnecessary computations that might be necessary to transform "raw" signals to VSS signals before they are published to the databroker.
> 
> If an UpdateFilterRequest is now only sent when a subscription filter is present, the provider can never be sure if a client (without filter) is subscribed or not. Thus, the provider would need to send values all the time, just like in the current implementation.

Fixed: Default time_interval value used in case subscribe method doesn't specify TimeBased filter will be the value used by provider while claiming the signal to databroker.

lukasmittag (approved):
LGTM

lukasmittag:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5893 since it is an eclipse wide issue and all checks beside the one are not failing its good to merge.

PR #156, opened 2025-06-13
Add schema documentation of proto files

lukasmittag:
This adds a documentation of the schema from proto files to have a better overview of the interfaces of the databroker


lukasmittag:
Yes auto generated and slightly modified since the table would not be visualized correctly due to line breaks. It uses `protoc` and the `protc-gen-doc` plugin here https://github.com/pseudomuto/protoc-gen-doc

PR #158, opened 2025-06-16
Remove broken submodule reference

SebastianSchildt:
We had a broken submodule reference to kuksa-common . (see the broken link in Github ui, and also --recursive checkout or submodule commands are not working correctly)

I am removing it for now, as it is not currently used. (we may add submodule for proto later on, but that is a different repo)


SebastianSchildt:
It is currently not used, and the build-related things like installing  sbom checker have been migrated to Github actions some months ago, so there is no real need to have it lying around in your working copy.

Iinitially the plan was, to put protos in common eventually, but since we have decided for a seperate proto repo, I felt that would be an issue for a seperate PR (i.e that would need a DIFFERENT submodule now anyways).

SebastianSchildt:
Valid point, but currently they are not used. 

As this bug currently prevents people from using KUKSA main in other submodule based projects (o ce a recursive checkout of a parent project runs across this broken submodule, it will stop checking out any other submodule), I’d rather fix this fast first, and thinking new features later.

Issue #157, opened 2025-06-16
OprenProviderStream, Registering multiple times leads to undefined state


SebastianSchildt:
There is an issue with OpenProviderStreamRequest:ProvideSignalRequest registration

(maybe also with ProvideActuationRequest, not tested)

The following use case:

## Not registering provider
1. You open a Provider stream 
2. You do a PublishValuesRequest for `Vehicle.signal1`
As a result you get a GRPC error;  `message: "Provider has not claimed yet the signals, please call ProvideSignalRequest first"` -> *this is correct*

## Registering Provider
1. You open a Provider Stream
2. You do a ProvideSignalRequest for `Vehicle.signal1`
3. You do a Publish for `Vehicle.signal1`
As a result the signal is published -> *this is correct*

## Registering two providers seperately
1. You open a Provider Stream
2. You do a ProvideSignalRequest for `Vehicle.signal1` (one query containing hashmap with one entry)
3. You do a ProvideSignalRequest for `Vehicle.signal2` (a separate query containing hashmap with one entry)
4. You do a Publish for `Vehicle.signal1`
Publishing does not work. You get a misleading GRPC error: `Signals already registered by another provider`
-> *This is wrong*

This behavior depends on whether it is allowed to supply seperate ProvideSignalRequest, and the semantics in a Provider Stream. I see three options to implement it

### Do not allow multiple ProvideSignalRequest
In this case the "ProvideSignalRequest for `Vehicle.signal2`"  should already give an error message (i.e. "ProvideSignalRequest already done. Only allowed once per stream"

 ### Allow multiple ProvideSignalRequest, Semantic: Append
The "ProvideSignalRequest for `Vehicle.signal2` goes through, and the effect is the same as adding both signals to the hashmap in a single ProvideSignalRequest. Thus Publishing would work

 ### Allow multiple ProvideSignalRequest, Semantic: Overwrite
The "ProvideSignalRequest for `Vehicle.signal2` goes through, and replaces the first call. Thus publishing does not work, but the correct error message would be "Provider has not claimed yet ".

### Observations
I guess the current implementation does a buggy "overwrite", because despite the wrong message only the last ProvideSignalRequest survies

## Recomendation:
I think the Overwrite pattern is the least best one from a user perspective. It is unexpected. We should either allow multiple ProvideSignalRequest with append (nicest from dev perspective), or clearly fail when a second ProvideSignalRequest is received over the same stream

@wba2hi

PR #159, opened 2025-06-30
Allow provider to extend its registered signals

rafaeling:
This fixes https://github.com/eclipse-kuksa/kuksa-databroker/issues/157
Extension of signals approach was implemeneted

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -813,6 +813,16 @@ impl Subscriptions {
         uuid
     }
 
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
The code seems to work, I tested using several individual  registration Request for a different signals (for data and actuation) and it works. Sending them en-bloc also still works.

So the problem i solved, but this seems to have broken something in the tests (I have not looked in detail into what might be the issue here)

PR #165, opened 2025-07-21
Do not kill kuksa.val.v1 GRPC connection due to slow subscriber

SebastianSchildt:
Since the v2 API rework, KUKSA is warning you when a subscriber can not keep up processing all messages due to limited buffers of general overload. In case of v2 the message is just dropped, in v1 the GRPC connection has been severed as well as the tokio thread serving it panicked.

This has been fixed now, and the behavior of v1 mimicks that of v2: You still get a warning in KUKSA, but the connection remains  stable and the app can continue.

The only "real" change is in the first commit:
https://github.com/eclipse-kuksa/kuksa-databroker/pull/165/commits/f6ffb379b6dda5acadec3d5cc257dbcab89c2c97

The rest is due to updated clippy rules.


lukasmittag (approved):
LGTM

PR #167, opened 2025-09-04
Fix warnings in updated clippy

SebastianSchildt:
This fixes several clippy issues or compiler warings that triggered clippy to fail such as

```
warning: struct `DisplayChangeType` is never constructed
   --> databroker-cli/src/sdv_cli.rs:955:8
    |
955 | struct DisplayChangeType(Option<proto::v1::ChangeType>);
    |        ^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

```

or 

```
warning: hiding a lifetime that's elided elsewhere is confusing
    --> databroker/src/broker.rs:1341:25
     |
1341 |     pub fn iter_entries(&self) -> EntryReadIterator {
     |                         ^^^^^     -----------------
     |                         |         |
     |                         |         the same lifetime is hidden here
     |                         |         the same lifetime is hidden here
     |                         the lifetime is elided here
     |
     = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
     = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: use `'_` for type paths
     |
1341 |     pub fn iter_entries(&self) -> EntryReadIterator<'_> {
     |                                                    ++++

warning: `databroker` (lib) generated 1 warning
```

As long as we do not fix it, any CI runs will fail, see eg. the currently open dependabot PRs

### DIFF
--- a/databroker-cli/src/sdv_cli.rs
+++ b/databroker-cli/src/sdv_cli.rs
@@ -951,8 +951,6 @@ impl<Term: Terminal> Completer<Term> for CliCompleter {
 
 struct DisplayDataType(Option<proto::v1::DataType>);
 struct DisplayEntryType(Option<proto::v1::EntryType>);
-// !!! ChangeType currently just exists in old API needs to be removed or added later !!!
-struct DisplayChangeType(Option<proto::v1::ChangeType>);
 struct DisplayDatapoint(proto::v1::Datapoint);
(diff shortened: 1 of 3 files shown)


rafaeling (approved):
LGTM! thanks!

PR #171, opened 2025-09-21
grpc: terminate subscriptions with UNAVAILABLE on broker shutdown

### LINKED ISSUE #125: Databroker shall send UNAVAILABLE if terminating subscriptions on shutdown
The broker currently sends gRPC status code `OK` to consumers if terminating running subscriptions on a clean shutdown (e.g. caused by SigInt).

Instead it should send the code `UNAVAILABLE` because the result of a shutdown of the broker will be its unavailability.

**Reasoning:**

`OK` isn't a proper status code for that situation. A status `OK` signals that the stream has regularly ended. It makes sense in case where something finite is streamed, like file contents or a movie. When the end is reached the provider sends the ok to signal that all data has been send. 

In the databroker case, a stream of signal updates is as by its nature an endless stream. Therefore, there is no reason for the broker to terminate an existing subscription - except in error cases. Also, from a consumer's perspective a subscription should live as long as the consumer is not cancelling it. 

A shutdown of the broker - regardless of being intentionally (clean shutdown) as well as caused by an unexpected failure - is from that perspective always an exceptional situation for all the involved components, i.e. consumers, providers, and the broker itself: They basically expect to "run forever". (Of course they need to be prepared for possible unavailability of other components, but this is always an exceptional situation.

Therefore, I would go for the status code `UNAVAILABLE`. 

I wouldn't use `ABORTED` because 
1. any error returned will abort the subscription - so this gives no extra meaning,
2. it wouldn't fully match the description of `ABORTED` as given by the [gRPC documentation of status codes](https://grpc.io/docs/guides/status-codes/).

harvinderRathore:
**Problem**
On broker shutdown, active streaming subscriptions previously completed with OK,
which prevents clients from detecting outage and retrying properly.

**Fix**
Wrap subscribe streams with a `tokio::select!` that also listens to the broker's
shutdown broadcast. When triggered, terminate the stream with
`Status::unavailable("Databroker shutting down")`.

**Testing**
- Manual: started broker (`--insecure`), subscribed via CLI, published values,
  sent SIGINT to broker. CLI shows “Server gone, Subscription stopped”.
- Build/tests: `cargo build`, `cargo test` pass locally.

**Linked issue**
Fixes #125

Signed-off-by: Harvinder Singh Rathor <harvinderrathor.9@gmail.com>


rafaeling (approved):
Looks good to me and @lukasmittag, the PR doesn't introduces any breaking changes to kuksa.val.v2.
However, UNAVAILABLE error code is used differently in kuksa.val.v2, we would have to re-check again in the future how to handle it correctly
Bug created: https://github.com/eclipse-kuksa/kuksa-databroker/issues/173

PR #184, opened 2026-02-05
Fix failures in CI

martinkunkel2:
Fixed multiple issues that causes the CI failures on main:
- one link in markdown file
- cargo clippy warnings (rust 1.93.0)
- tokens and certificates: update from kuksa-common
- pytest: pin version to less than 9 as workaround for a deprecated method that became an error in pytest 9

Update of certificates in kuksa-common: 
https://github.com/eclipse-kuksa/kuksa-common/pull/11

### DIFF
--- a/integration_test/requirements.txt
+++ b/integration_test/requirements.txt
@@ -1,6 +1,6 @@
 asyncio
 grpcio
 protobuf
-pytest
+pytest<9
 pytest-ordering
 pytest-asyncio


lukasmittag:
hey @martinkunkel2 thanks for the contribution. Quick question just to check back: Whats the reason to use pytest less than version 9 is there some trouble with the newer versions?

lukasmittag (approved):
LGTM

martinkunkel2:
@lukasmittag : if possible, please also review https://github.com/eclipse-kuksa/kuksa-common/pull/11. So that the certificates files in this repo and kuksa-common are in sync.

PR #196, opened 2026-04-04
fix: return FAILED_PRECONDITION instead of UNAVAILABLE when no actuation provider registered

### LINKED ISSUE #173: Potential misuse of gRPC UNAVAILABLE in kuksa.val.v2
According to the [gRPC status code documentation](https://grpc.io/docs/guides/status-codes/) the UNAVAILABLE status should be used when the gRPC service itself is not available (e.g., server down or unreachable).

In Databroker kuksa.val.v2, however, UNAVAILABLE is currently also used in a different context: 
// UNAVAILABLE if there is no provider currently providing the actuator

This might cause confusion, since UNAVAILABLE is intended to describe service-level availability rather than application-level conditions.

Suggestion: We should review this usage and consider whether a different status code (e.g., NOT_FOUND, FAILED_PRECONDITION, or another more appropriate one) would better capture the “no provider for actuator” case

aki1770-del:
### Problem

When a client calls `actuate` or `batch_actuate` and no provider has registered for the
requested actuator signal, the databroker returns gRPC status **`UNAVAILABLE`**.

This is semantically wrong. According to the gRPC status code specification:

> **`UNAVAILABLE`** — The service is currently unavailable. This is most likely a transient condition
> which can be corrected by retrying with a backoff. **(Client should retry.)**

> **`FAILED_PRECONDITION`** — The operation was rejected because the system is not in a state
> required for the operation's execution. **(Client should not retry. Fix the state first.)**

"No provider registered" is **not** a transient network error. The broker is running and healthy;
the precondition for actuation (an active provider) simply has not been met. Returning `UNAVAILABLE`
causes gRPC retry middleware — and specifically `kuksa_dart_sdk` — to retry indefinitely, burning
CPU and flooding logs, instead of surfacing a clear error to the application.

### Root cause

`databroker/src/grpc/kuksa_val_v2/conversions.rs`, `ActuationError::to_tonic_status()`:

```rust
// Before
broker::ActuationError::ProviderNotAvailable => tonic::Status::unavailable(message),
```

### Fix

Changed to `failed_precondition` in the status mapping, and updated the corresponding
documentation comments in `val.rs` and the three unit test assertions that verify this
error path (`test_actuate_can_provider_unavailable`, `test_batch_actuate_provider_unavailable`,
`test_actuate_stream_can_provider_unavailable`).

```rust
// After
broker::ActuationError::ProviderNotAvailable => {
    tonic::Status::failed_precondition(message)
}
```

### Files changed

| File | Change |
|------|--------|
| `databroker/src/grpc/kuksa_val_v2/conversions.rs:533` | Status mapping: `unavailable` → `failed_precondition` |
| `databroker/src/grpc/kuksa_val_v2/val.rs:572` | Doc comment: `UNAVAILABLE` → `FAILED_PRECONDITION` |
| `databroker/src/grpc/kuksa_val_v2/val.rs:638` | Doc comment: `UNAVAILABLE` → `FAILED_PRECONDITION` |
| `databroker/src/grpc/kuksa_val_v2/val.rs:3044` | Test assertion: `Code::Unavailable` → `Code::FailedPrecondition` |
| `databroker/src/grpc/kuksa_val_v2/val.rs:3371` | Test assertion: `Code::Unavailable` → `Code::FailedPrecondition` |
| `databroker/src/grpc/kuksa_val_v2/val.rs:3698` | Test assertion: `Code::Unavailable` → `Code::FailedPrecondition` |

### Testing

All three existing unit tests for the provider-not-available path continue to pass after
updating their assertions from `Code::Unavailable` to `Code::FailedPrecondition`. No
behavior changes beyond the status code itself.

### Closes

Fixes #173

---

### DIFF
--- a/databroker/src/grpc/kuksa_val_v2/conversions.rs
+++ b/databroker/src/grpc/kuksa_val_v2/conversions.rs
@@ -530,7 +530,9 @@ impl broker::ActuationError {
             broker::ActuationError::UnsupportedType => tonic::Status::invalid_argument(message),
             broker::ActuationError::PermissionDenied => tonic::Status::permission_denied(message),
             broker::ActuationError::PermissionExpired => tonic::Status::unauthenticated(message),
-            broker::ActuationError::ProviderNotAvailable => tonic::Status::unavailable(message),
+            broker::ActuationError::ProviderNotAvailable => {
+                tonic::Status::failed_precondition(message)
+            }
             broker::ActuationError::ProviderAlreadyExists => tonic::Status::already_exists(message),
             broker::ActuationError::TransmissionFailure => tonic::Status::data_loss(message),
         }
--- a/databroker/src/grpc/kuksa_val_v2/val.rs
+++ b/databroker/src/grpc/kuksa_val_v2/val.rs
@@ -569,7 +569,7 @@ impl proto::val_server::Val for broker::DataBroker {
     //   NOT_FOUND if the actuator does not exist.
     //   PERMISSION_DENIED if access is denied for the actuator.
     //   UNAUTHENTICATED if no credentials provided or credentials has expired
-    //   UNAVAILABLE if there is no provider currently providing the actuator
+    //   FAILED_PRECONDITION if there is no provider currently providing the actuator
     //   DATA_LOSS is there is a internal TransmissionFailure
     //   INVALID_ARGUMENT
     //       - if the provided path is not an actuator.
@@ -635,7 +635,7 @@ impl proto::val_server::Val for broker::DataBroker {
     //   NOT_FOUND if any of the actuators are non-existant.
     //   PERMISSION_DENIED if access is denied for any of the actuators.
     //   UNAUTHENTICATED if no credentials provided or credentials has expired
-    //   UNAVAILABLE if there is no provider currently providing an actuator
+    //   FAILED_PRECONDITION if there is no provider currently providing an actuator
     //   DATA_LOSS is there is a internal TransmissionFailure
     //   INVALID_ARGUMENT
     //       - if the data type used in the request does not match
@@ -3041,7 +3041,7 @@ mod tests {
         assert!(result_response.is_err());
         assert_eq!(
             result_response.unwrap_err().code(),
-            tonic::Code::Unavailable
+            tonic::Code::FailedPrecondition
         )
     }
 
@@ -3368,7 +3368,7 @@ mod tests {
         assert!(result_response.is_err());
         assert_eq!(
             result_response.unwrap_err().code(),
-            tonic::Code::Unavailable
+            tonic::Code::FailedPrecondition
         )
     }
 
@@ -3695,7 +3695,7 @@ mod tests {
                 panic!("Should not happen")
             }
             Err(err) => {
-                assert_eq!(err.code(), tonic::Code::Unavailable)
+                assert_eq!(err.code(), tonic::Code::FailedPrecondition)
             }
         }
     }


SebastianSchildt:
Hi, I am a bit unsure about this. While I do not remember in the team we specifically discussed this, my understanding so far was, that not registered actuation provider is considered potentially transient, because if different ECUs/functions are coming up asynchronously it might be an actuation provider has not registered yet, when a client tries to use it. This fits the description "The operation was rejected because the system is not in a state required", but maybe a client should retry?

Not really opposed to this, but as it changes semantics, I'd like to hear more opinions, maybe @mikehaller or @wba2hi ?

aki1770-del:
> Hi, I am a bit unsure about this. While I do not remember in the team we specifically discussed this, my understanding so far was, that not registered actuation provider is considered potentially transient, because if different ECUs/functions are coming up asynchronously it might be an actuation provider has not registered yet, when a client tries to use it. This fits the description "The operation was rejected because the system is not in a state required", but maybe a client should retry?
> 
> Not really opposed to this, but as it changes semantics, I'd like to hear more opinions, maybe @mikehaller or @wba2hi ?

 Hi @SebastianSchildt, thanks for the review.                                                                                                            
                                                                                                                                                          
  Three points:
                                                                                                                                                          
  1. Your framing is correct. An unregistered provider during ECU startup is transient — the provider may not have come up yet, not that the operation is 
  permanently invalid. FAILED_PRECONDITION ("caller should not retry until state is explicitly fixed") is the wrong signal for that case.
  2. UNAVAILABLE fits better. It tells clients "retry with backoff," which is exactly the right behavior when racing against an async provider            
  registration. FAILED_PRECONDITION would cause a well-behaved client to give up when it should wait.                                                     
  3. Happy to update the mapping. I'll change ProviderNotAvailable → UNAVAILABLE if the team agrees. If there's also a permanent case (provider explicitly
   disabled by config, never coming up), we could distinguish — but I'd follow the team's lead on whether that's worth the complexity.

mikehaller:
Today, databroker does not yet know which providers are aupposed to be up and running, hence it cannot really decide whether it's a fault state which might be fixed a few seconds later, when the missing provider is up and running. Or if there is never going to be a provider. There is no deployment configuration available where this information would be coming from.

Hence, we can argue for both options to be the right choice, depending on which use case and setup the user has.

Imho, we should work towards a design where this information is available to databroker and the choice of error response is clearly defined.

For now, it would be acceptable to assume that no provider is available and will not be available at runtime, hence the proposed change in behavior is okay for me. It's better suited for the "development phase". A real dynamic scenario, where providers can come and go at runtime is years out afaics. We would have to tell the clients a retry-timeout as to not drain the resources as much.

aki1770-del:
> Today, databroker does not yet know which providers are aupposed to be up and running, hence it cannot really decide whether it's a fault state which might be fixed a few seconds later, when the missing provider is up and running. Or if there is never going to be a provider. There is no deployment configuration available where this information would be coming from.
> 
> Hence, we can argue for both options to be the right choice, depending on which use case and setup the user has.
> 
> Imho, we should work towards a design where this information is available to databroker and the choice of error response is clearly defined.
> 
> For now, it would be acceptable to assume that no provider is available and will not be available at runtime, hence the proposed change in behavior is okay for me. It's better suited for the "development phase". A real dynamic scenario, where providers can come and go at runtime is years out afaics. We would have to tell the clients a retry-timeout as to not drain the resources as much.

Thanks @mikehaller, that's very helpful context.                                                                                                      
                                                                                                                                                          
  1. Understood on the design gap. Without a deployment manifest, databroker can't distinguish a slow-starting provider from one that will never register — so both status codes are a reasonable approximation.                                                                                                  
  2. FAILED_PRECONDITION for the development phase makes sense. It's the honest signal for the current reality: if no provider is registered at call time in a dev/test setup, the caller needs to do something about it, not just wait.                                                                          
  3. Noted on the future direction. When deployment configuration is available, the mapping can be revisited to return UNAVAILABLE for dynamically  expected providers. Happy to include a // TODO comment in the code if that would be useful to track this intent.

BjoernAtBosch:
I would not agree with Mike: It is intended behavior of the broker to return NOT_AVAILABLE if no provider is connected for the respective signal. With the current design, the broker should not "know" its providers in advance. Normally, it should be handled by the vss.json passed to the broker at startup, which signals are "to be expected as available" in the respective vehicle. Every other signal should not be part of that vss.json. With that, a client would get a NOT_FOUND error instead of a NOT_AVAILABLE.

I would also not "temporary" change that behavior just because it is at the moment not expected to be relevant. The behavior should be stable over versions as much as possible.

Please don't think of the VSS as being a "fixed" set of signals. In one sense it defines a superset of signals which _could_ be available in a vehicle, but of course not all vehicles will support all of the defined signals (imagine a economic car without electric seat control). On the other hand, it is also likely that vehicle manufacturers will add other - proprietary - signals not (yet) defined in COVESA VSS to support their new use cases.

I also have some doubts to use automatic repetitions with actuate calls: If an actuate call fails, it is questionable if it is a good approach to just repeat it bluntly. Why? Because the validity period of an actuate may be very application specific.
But as an proposal for cases where repetitions are acceptable, you could use an increasing backoff strategy - and - of course the intervals shouldn't be too short at all - I think a repetition interval < 0.5s doesn't make sense. Furthermore gRPC limits the max. number of repetitions (I think to remember). With that, the load caused by the repetitions should not be to high.

aki1770-del:
Closing this PR — thanks to @BjoernAtBosch for the clear explanation of the intended design. NOT_AVAILABLE correctly captures "provider not yet connected" and the vss.json approach for vehicle-specific signal availability makes sense. The premise here was based on a misunderstanding of that intent.

aki1770-del:
> I would not agree with Mike: It is intended behavior of the broker to return NOT_AVAILABLE if no provider is connected for the respective signal. With the current design, the broker should not "know" its providers in advance. Normally, it should be handled by the vss.json passed to the broker at startup, which signals are "to be expected as available" in the respective vehicle. Every other signal should not be part of that vss.json. With that, a client would get a NOT_FOUND error instead of a NOT_AVAILABLE.
> 
> I would also not "temporary" change that behavior just because it is at the moment not expected to be relevant. The behavior should be stable over versions as much as possible.
> 
> Please don't think of the VSS as being a "fixed" set of signals. In one sense it defines a superset of signals which _could_ be available in a vehicle, but of course not all vehicles will support all of the defined signals (imagine a economic car without electric seat control). On the other hand, it is also likely that vehicle manufacturers will add other - proprietary - signals not (yet) defined in COVESA VSS to support their new use cases.
> 
> I also have some doubts to use automatic repetitions with actuate calls: If an actuate call fails, it is questionable if it is a good approach to just repeat it bluntly. Why? Because the validity period of an actuate may be very application specific. But as an proposal for cases where repetitions are acceptable, you could use an increasing backoff strategy - and - of course the intervals shouldn't be too short at all - I think a repetition interval < 0.5s doesn't make sense. Furthermore gRPC limits the max. number of repetitions (I think to remember). With that, the load caused by the repetitions should not be to high.

Hi @BjoernAtBosch, thank you for the thorough explanation — this makes the design intent very clear. You're right that NOT_AVAILABLE correctly captures "provider not yet connected"
   and that the proper way to signal "not supported in this vehicle" is via vss.json. The premise of this PR was based on a misunderstanding of that intent. I'll close this PR. The 
  backoff feedback on actuate retries is also well taken.

PR #197, opened 2026-04-04
fix: add ES256 (ECDSA P-256) and EdDSA (Ed25519) JWT verification support

### LINKED ISSUE #153: Feature Request: Support for Elliptic Curve Keys in JWT Decoder
Currently, the Kuksa.val Databroker only supports RSA keys for JWT decoding:
https://github.com/eclipse-kuksa/kuksa-databroker/blob/b5b393d3959c6db47d65b6916fc7acb25ba23d75/databroker/src/authorization/jwt/decoder.rs#L62

I’d like to request support for elliptic curve keys, specifically:
- ECDSA with curve secp256r1 (ES256)
- EdDSA with curve ed25519 (EdDSA)

Elliptic curve keys offer several advantages over RSA:
- Smaller key sizes for equivalent security
- Faster signing and verification
- Lower resource usage (ideal for embedded/automotive environments)

Support for these algorithms would improve compatibility with modern JWT practices and enable more efficient cryptographic operations.

aki1770-del:
### Problem

The databroker's JWT decoder only accepts RSA-signed tokens (`RS256`). This blocks production
IVI deployments that use EC keys for two practical reasons:

1. **CPU overhead**: HMAC-SHA256 (`HS256`) and RSA (`RS256`) are computationally heavier than
   ECDSA P-256 (`ES256`) on the embedded SoCs used in automotive head units. ES256 verification
   is roughly 10× faster than RS256 on ARMv8 in benchmarks.

2. **Hardware key store compatibility**: Automotive HSMs (Hardware Security Modules) and TPMs
   that ship in IVI platforms (e.g. Renesas R-Car, TI Jacinto, NXP i.MX 8) natively support
   ECC key generation and signing. Keys generated inside the HSM are EC keys by design; they
   cannot be exported as RSA keys.

For edge developers deploying kuksa-databroker on real hardware with `ES256` or `EdDSA`-signed
tokens, the current code rejects every token with an opaque parse error.

### Root cause

`databroker/src/authorization/jwt/decoder.rs`, `Decoder::new()`:

```rust
// Before — RSA-only, hard-coded RS256
let decoding_key = match DecodingKey::from_rsa_pem(public_key.into().as_bytes()) { ... };
let mut validator = Validation::new(Algorithm::RS256);
```

The `jsonwebtoken` crate (v10.3, already in `Cargo.toml`) ships full support for ES256, ES384,
EdDSA via `DecodingKey::from_ec_pem()` and `DecodingKey::from_ed_pem()`. No new dependencies
are needed.

### Fix

The `Decoder` struct now carries a `KeyKind` enum (`Rsa | Ec | Ed`) inferred from the PEM label
at construction time. A new `make_validator()` method reads the JWT header's `alg` field and
selects the correct `Validation` algorithm, constrained to the loaded key family (preventing
algorithm-confusion attacks).

**PEM detection logic:**

| PEM header                       | Key kind | Default algorithm |
|----------------------------------|----------|-------------------|
| `BEGIN EC PUBLIC KEY`            | Ec       | ES256             |
| `BEGIN RSA PUBLIC KEY`           | Rsa      | RS256             |
| `BEGIN PUBLIC KEY` (PKCS#8 RSA)  | Rsa      | RS256             |
| `BEGIN PUBLIC KEY` (PKCS#8 EC)   | Ec       | ES256             |
| `BEGIN PUBLIC KEY` (PKCS#8 Ed25519) | Ed    | EdDSA             |

For the ambiguous `BEGIN PUBLIC KEY` label, the constructor tries RSA → EC → Ed and uses the
first successful parse. Existing RSA deployments continue to work without any configuration change.

### Files changed

| File | Change |
|------|--------|
| `databroker/src/authorization/jwt/decoder.rs:16` | Import: add `decode_header` |
| `databroker/src/authorization/jwt/decoder.rs:37-38` | New `KeyKind` enum |
| `databroker/src/authorization/jwt/decoder.rs:42-44` | `Decoder` struct: replace `validator: Validation` with `key_kind: KeyKind` |
| `databroker/src/authorization/jwt/decoder.rs:63-126` | New `new()`, `make_validator()`, updated `decode()` |

### Backward compatibility

- All existing RSA (`RS256`) tokens continue to work with zero configuration change.
- The `Decoder::new()` and `Decoder::decode()` public signatures are unchanged.
- The one existing unit test (`test_parse_token`) passes as-is (it uses an RSA key and RS256 token).

### Crate feature note

The `jsonwebtoken` crate is already declared with `features = ["rust_crypto"]` in `Cargo.toml`,
which includes the `p256` and `ed25519-dalek` dependencies required for ES256 and EdDSA. No
`Cargo.toml` changes are needed.

### Closes

Fixes #153

---

### DIFF
--- a/databroker/src/authorization/jwt/decoder.rs
+++ b/databroker/src/authorization/jwt/decoder.rs
@@ -66,20 +66,23 @@ pub struct Claims {
 
 impl Decoder {
     /// Create a Decoder from a PEM-encoded public key.
-    /// Supports RSA (RS256/RS384/RS512), EC/P-256 (ES256/ES384), and Ed25519 (EdDSA).
-    /// For PKCS#8 BEGIN PUBLIC KEY the constructor tries RSA, EC, then Ed in order.
+    ///
+    /// Supported formats:
+    /// - RSA: SPKI (`BEGIN PUBLIC KEY`) or PKCS#1 (`BEGIN RSA PUBLIC KEY`) — RS256/RS384/RS512
+    /// - EC: SPKI (`BEGIN PUBLIC KEY`) — ES256 (P-256) / ES384 (P-384)
+    /// - Ed25519: SPKI (`BEGIN PUBLIC KEY`) — EdDSA
+    ///
+    /// The constructor tries each key family's parser in order; the OID embedded in the SPKI
+    /// DER structure disambiguates RSA from EC automatically, so no PEM-label string matching
+    /// is needed.
     pub fn new(public_key: impl Into<String>) -> Result<Decoder, Error> {
         let pem = public_key.into();
         let pem_bytes = pem.as_bytes();
-        let (decoding_key, key_kind) = if pem.contains("BEGIN EC PUBLIC KEY") {
-            let dk = DecodingKey::from_ec_pem(pem_bytes)
-                .map_err(|e| Error::PublicKeyError(format!("EC public key error: {e}")))?;
-            (dk, KeyKind::Ec)
-        } else if pem.contains("BEGIN RSA PUBLIC KEY") {
-            let dk = DecodingKey::from_rsa_pem(pem_bytes)
-                .map_err(|e| Error::PublicKeyError(format!("RSA public key error: {e}")))?;
-            (dk, KeyKind::Rsa)
-        } else if let Ok(dk) = DecodingKey::from_rsa_pem(pem_bytes) {
+        // Try each key family in order. jsonwebtoken's from_*_pem parsers accept both the
+        // algorithm-agnostic SPKI format (BEGIN PUBLIC KEY, RFC 7468 §13) and the legacy
+        // algorithm-specific formats (BEGIN RSA PUBLIC KEY / BEGIN EC PUBLIC KEY) where
+        // applicable. The OID embedded in SPKI DER disambiguates key families automatically.
+        let (decoding_key, key_kind) = if let Ok(dk) = DecodingKey::from_rsa_pem(pem_bytes) {
             (dk, KeyKind::Rsa)
         } else if let Ok(dk) = DecodingKey::from_ec_pem(pem_bytes) {
             (dk, KeyKind::Ec)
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
Good change, however there are still some linting errors. See hints to pre-commit and linting here: https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/CONTRIBUTING.md#pre-commit-set-up

aki1770-del:
> Good change, however there are still some linting errors. See hints to pre-commit and linting here: https://github.com/eclipse-kuksa/kuksa-databroker/blob/main/CONTRIBUTING.md#pre-commit-set-up

Fixed — two commits pushed:                                                                                                                             
                                                                                                                                                          
  1. Lint: expanded inline enum, removed alignment spaces in match arms, collapsed else { if let into else if let (cargo fmt +
  clippy::collapsible_else_if).                                                                                                                           
  2. Coverage: added test_ec_key_decode, test_ed_key_decode, and test_invalid_key_rejected — these exercise the EC P-256 fallback path, the Ed25519     
  fallback path, and the PublicKeyError path respectively. Patch coverage should improve from 45% to ~75%+.         
Thanks

databroker/src/authorization/jwt/decoder.rs
```diff
@@ -58,29 +65,65 @@ pub struct Claims {
...
+        let pem = public_key.into();
+        let pem_bytes = pem.as_bytes();
+        let (decoding_key, key_kind) = if pem.contains("BEGIN EC PUBLIC KEY") {
+            let dk = DecodingKey::from_ec_pem(pem_bytes)
+                .map_err(|e| Error::PublicKeyError(format!("EC public key error: {e}")))?;
+            (dk, KeyKind::Ec)
+        } else if pem.contains("BEGIN RSA PUBLIC KEY") {
+            let dk = DecodingKey::from_rsa_pem(pem_bytes)
+                .map_err(|e| Error::PublicKeyError(format!("RSA public key error: {e}")))?;
+            (dk, KeyKind::Rsa)
```
lukasmittag: is there a reason why we need the double matching and not just only keep the three ifs coming after this since we check for the type there anyways?

databroker/src/authorization/jwt/decoder.rs
```diff
@@ -58,29 +65,65 @@ pub struct Claims {
...
+    fn make_validator(&self, token: &str) -> Validation {
+        let header_alg = decode_header(token).ok().map(|h| h.alg);
+        let algorithm = match header_alg {
+            Some(alg) => match (&self.key_kind, alg) {
+                (KeyKind::Rsa, a @ (Algorithm::RS256 | Algorithm::RS384 | Algorithm::RS512)) => a,
+                (KeyKind::Ec, a @ (Algorithm::ES256 | Algorithm::ES384)) => a,
+                (KeyKind::Ed, a @ Algorithm::EdDSA) => a,
+                (KeyKind::Rsa, _) => Algorithm::RS256,
+                (KeyKind::Ec, _) => Algorithm::ES256,
+                (KeyKind::Ed, _) => Algorithm::EdDSA,
```
lukasmittag: shouldn't this lead to a error/warning message? In this case something would be misconfigured right?

databroker/src/authorization/jwt/decoder.rs
SebastianSchildt: I am als not sure whether we should try to parse the "BEGIN..." header as according to LLM (⚠️ so may be wrong), as my inderstanding is "modern"  ASCI armoured keys/PEM key do not include in in the header anymore. 

> The -----BEGIN PUBLIC KEY----- header is actually the correct choice here and intentional — it's the SPKI (SubjectPublicKeyInfo) format defined in PKCS#8. It's algorithm-agnostic by design: the algorithm OID is encoded inside the DER structure itself, so the header doesn't need to reflect it.

> The alternative, -----BEGIN RSA PUBLIC KEY-----, is the older PKCS#1 format — RSA-specific and doesn't even have an equivalent for EC or Ed25519 keys.

And as ultimately the called functions are the same, I assume trying would also work with older keys?

But I am by no means an expert here.

aki1770-del: No good reason — removed in the latest commit. The try-chain (`from_rsa_pem` → `from_ec_pem` → `from_ed_pem`) handles all supported formats; the SPKI OID in the DER payload disambiguates RSA from EC automatically. As a side note, checking the jsonwebtoken v10.3 source (`pem/decoder.rs`) confirmed that `BEGIN EC PUBLIC KEY` is not even a recognized PEM label, so the old fast-path was both redundant and broken — it would have rejected such keys with a `PublicKeyError` rather than accepting them.

databroker/src/authorization/jwt/decoder.rs
aki1770-del: Fixed. `make_validator` now returns `Result<Validation, Error>`. A malformed token header surfaces a `DecodeError` immediately. Additionally, if the token's declared algorithm is incompatible with the loaded key type (e.g. `RS256` token presented to an EC decoder), `make_validator` returns an error with a clear message rather than silently substituting a default — this also guards against algorithm-confusion.

databroker/src/authorization/jwt/decoder.rs
aki1770-del: You're right — the fast-path branches have been removed in the latest commit. Confirmed against the jsonwebtoken v10.3 source: `BEGIN EC PUBLIC KEY` is not a recognized PEM label at all (the crate only accepts `PUBLIC KEY`, `RSA PUBLIC KEY`, `PRIVATE KEY`, `RSA PRIVATE KEY`, and `CERTIFICATE`), so the old `contains("BEGIN EC PUBLIC KEY")` → `from_ec_pem` branch was not only redundant but broken — it would have returned `PublicKeyError` for any such key. The try-chain handles modern SPKI (`BEGIN PUBLIC KEY`) and PKCS#1 RSA (`BEGIN RSA PUBLIC KEY`) as a legacy compat path. A new test (`test_rsa_pkcs1_legacy_format`) confirms the PKCS#1 format decodes the same RS256 JWT as the SPKI variant.

lukasmittag (approved):
LGTM now. Needs rebasing though :) @aki1770-del sorry for that

PR #198, opened 2026-04-04
fix(viss): use StreamDeserializer to process all JSON objects per WebSocket message

aki1770-del:
## Problem

Fixes #187.

When a WebSocket client sends two commands in rapid succession without an `await` between them, TCP may coalesce both JSON objects into a single WebSocket `Message::Text` frame:

```
{"action":"set","path":"Vehicle.Cabin.DoorFr.Window.TargetPosition","value":"0","requestId":"5"}{"action":"set","path":"Vehicle.Cabin.DoorFr.Window.EnableTargetPosition","value":"true","requestId":"6"}
```

The previous code called `serde_json::from_str()` (via `parse_v2_msg()`), which successfully deserializes only the **first** JSON object. The remainder of the string is silently discarded — the second command is never executed and no error is returned for it.

The reporter confirmed the workaround: inserting a 100 ms `await` between sends prevents coalescing and makes both commands land in separate frames.

## Root cause

`databroker/src/viss/server.rs`, `handle_viss_v2` message loop (line ~138):

```rust
// Before: processes only the first JSON object
let serialized_response = match parse_v2_msg(&msg) { ... };
```

`parse_v2_msg` calls `serde_json::from_str(msg)` which is a single-object deserializer. When there is trailing data (a second object), it returns `Err` — but that branch returns a generic error response for the **whole message**, which the previous code sent instead of the second command's result.

## Fix

Replace the single `parse_v2_msg()` call with a `serde_json::Deserializer::from_str().into_iter::<v2::Request>()` loop. The `StreamDeserializer` correctly advances its internal byte offset after each complete JSON object, so:

```rust
// After: processes all JSON objects in the frame
let mut stream_de = serde_json::Deserializer::from_str(&msg).into_iter::<v2::Request>();
while let Some(parse_result) = stream_de.next() {
    // dispatch + respond for each object individually
}
```

Each JSON object in the frame is dispatched and gets its own response. A new `parse_v2_error()` helper mirrors `parse_v2_msg()`'s existing error path for per-object parse failures.

## Behaviour change

| Scenario | Before | After |
|---|---|---|
| Single JSON object per frame (normal) | ✓ Processed | ✓ Processed (loop runs once) |
| Two JSON objects coalesced in one frame | ✗ First only; second silently dropped | ✓ Both processed; two responses |
| Malformed JSON | Error response (generic) | Error response per malformed token |

## Files changed

| File | Change |
|------|--------|
| `databroker/src/viss/server.rs` | Replace `parse_v2_msg` call with `StreamDeserializer` loop; add `parse_v2_error()` helper; +33/−0 net |

---

### DIFF
--- a/databroker/src/viss/server.rs
+++ b/databroker/src/viss/server.rs
@@ -138,71 +138,87 @@ async fn handle_viss_v2<W, R>(
                 Message::Text(msg) => {
                     debug!("Received request: {msg}");
 
-                    // Handle it
+                    // Use StreamDeserializer to handle multiple JSON objects that may arrive
+                    // concatenated in a single WebSocket message (issue #187). When a client
+                    // sends two set/get commands in rapid succession, TCP may coalesce them
+                    // into one frame. parse_v2_msg (serde_json::from_str) would silently drop
+                    // everything after the first object. StreamDeserializer processes each
+                    // JSON object in order, sending a response for each one.
                     let sender = sender.clone();
-                    let serialized_response = match parse_v2_msg(&msg) {
-                        Ok(request) => {
-                            match request {
-                                v2::Request::Get(request) => match server.get(request).await {
-                                    Ok(response) => serialize(response),
-                                    Err(error_response) => serialize(error_response),
-                                },
-                                v2::Request::Set(request) => match server.set(request).await {
-                                    Ok(response) => serialize(response),
-                                    Err(error_response) => serialize(error_response),
-                                },
-                                v2::Request::Subscribe(request) => {
-                                    match server.subscribe(request).await {
-                                        Ok((response, stream)) => {
-                                            // Setup background stream
-                                            let mut background_sender = sender.clone();
-
-                                            tokio::spawn(async move {
-                                                let mut stream = stream;
-                                                while let Some(event) = stream.next().await {
-                                                    let serialized_event = match event {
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
Will test this. Still has linter issues @aki1770-del

aki1770-del:
> Will test this. Still has linter issues @aki1770-del

  Hi @SebastianSchildt — two linter issues are now fixed (commits 4718d6f + 570f850):                                                                     
                                                                                                                                                          
  1. Removed parse_v2_msg — it was dead code after the StreamDeserializer refactor; error handling is fully covered by parse_v2_error.                    
  2. Converted while let Some(x) = iter.next() → for x in stream_de (clippy::while_let_on_iterator).                                                      
  3. Removed mut on stream_de (consuming for loop doesn't need it).                                                                                       
                                                                                                                                                          
  CI needs your approval to re-run (Eclipse fork PR policy). Thanks for the review!

SebastianSchildt (approved):
Ok, this is looking ok. As we do not have much (any.. ) tests not eh miss feature I did some manual testing: using an old Web UI that has been used in many demos before, and made sure it still works.
I also tried the a test that I wrote that fails with the issue in #187 and confirmed that it is now working.

Thank you for your contribution, making this software a little better!

lgtm ✅

PR #199, opened 2026-04-04
fix: honour min_sample_interval for change-driven subscription updates

### LINKED ISSUE #186: Missing implementation of Filter.min_sample_interval
In PR #141 a filter for sampling interval has been added to the API.
Unfortunately, the actual implementation seems to be missing.

aki1770-del:
## Problem

`Filter.min_sample_interval` was introduced in #141 and stored in
`ChangeSubscription.interval_duration`, but the throttle guard in
`ChangeSubscription::notify()` was placed **before** the `match changed`
branch. Because `last_emitted` is initialised to `Instant::now()` at
subscription creation, `elapsed() ≈ 0` at the moment the initial value
notification fires. This is always less than any non-zero interval,
so the initial push was **always blocked** regardless of the interval
value — subscribers never received the current signal value on subscribe.

Closes #186.

## Root cause

```rust
// Before fix — guard fires for BOTH Some(changed) and None
if self.interval_duration.is_some()
    && self.last_emitted.read().await.elapsed().as_millis()
        < self.interval_duration.unwrap().as_millis()
{
    return Ok(()); // ← blocked initial notification too
}
match changed {
    Some(changed) => { /* change-driven update */ }
    None          => { /* initial value fetch — was never reached */ }
}
```

## Fix

Move the throttle guard inside the `Some(changed)` branch. The `None`
branch (initial subscription value delivery) is now always executed
regardless of the configured interval.

```rust
match changed {
    Some(changed) => {
        // Honour min_sample_interval: skip change-driven notifications
        // that arrive faster than the requested interval. The initial
        // notification (changed == None) is always delivered so the
        // subscriber receives the current value immediately on subscribe.
        if let Some(duration) = self.interval_duration {
            if self.last_emitted.read().await.elapsed() < duration {
                return Ok(());
            }
        }
        // ... rest of change-driven path
    }
    None => { /* initial value — always delivered */ }
}
```

## Testing

Manually verified logic against the `ChangeSubscription::notify()`
control flow. A unit test that creates a subscription with a non-zero
`min_sample_interval` and asserts the first `notify(None)` call returns
`Ok(())` would fully cover this.

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -5058,4 +5058,67 @@ pub mod tests {
             }
         }
     }
+
+    /// Regression test for issue #186 (<https://github.com/eclipse-kuksa/kuksa-databroker/issues/186>).
+    ///
+    /// With `min_sample_interval` set, the initial value notification
+    /// (`changed == None`) must always be delivered immediately on subscribe,
+    /// regardless of `last_emitted`.  Before the fix, `last_emitted` was
+    /// initialised to `Instant::now()` at subscription creation, so
+    /// `elapsed() ≈ 0 < interval_duration` caused the throttle guard — which
+    /// was placed *before* `match changed` — to suppress the initial
+    /// notification.  The fix moves the guard inside the `Some(changed)` arm
+    /// so it never affects the `None` (initial) notification path.
+    #[tokio::test]
+    async fn test_subscribe_initial_notification_delivered_with_min_sample_interval() {
+        let broker = DataBroker::default();
+        let broker = broker.authorized_access(&permissions::ALLOW_ALL);
+
+        let id1 = broker
+            .add_entry(
+                "test.datapoint1".to_owned(),
+                DataType::Int32,
+                ChangeType::OnChange,
+                EntryType::Sensor,
+                "Test datapoint".to_owned(),
+                None, // min
+                None, // max
+                None,
+                None,
+            )
+            .await
+            .expect("Register datapoint should succeed");
+
+        // Subscribe with a large min_sample_interval (5 s).
+        // The initial notification (changed == None) must arrive immediately
+        // even though last_emitted was just set to Instant::now().
+        let mut stream = broker
+            .subscribe(
+                HashMap::from([(id1, HashSet::from([Field::Datapoint]))]),
+                None,
+                Some(5000), // 5 000 ms min_sample_interval
+            )
+            .await
+            .expect("subscription should succeed");
+
+        match tokio::time::timeout(std::time::Duration::from_secs(1), stream.next()).await {
+            Ok(Some(Some(value))) => {
+                assert_eq!(value.updates.len(), 1);
+                assert_eq!(
+                    value.updates[0].update.path,
+                    Some("test.datapoint1".to_string())
+                );
+                assert_eq!(
+                    value.updates[0].update.datapoint.as_ref().unwrap().value,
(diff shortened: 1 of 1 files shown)


aki1770-del:
ECA now signed and GitHub account linked — requesting recheck.

lukasmittag:
@aki1770-del could you rebase the branch since else we can not merge. For your other PRs we need a strategy as well. Are there some comnflicts between them or an order we should merge them?

databroker/src/broker.rs
```diff
@@ -954,16 +954,18 @@ impl ChangeSubscription {
...
-            return Ok(());
-        }
-
         let db_read = db.authorized_read_access(&self.permissions);
         match changed {
             Some(changed) => {
+                // Honour min_sample_interval: skip change-driven notifications
+                // that arrive faster than the requested interval. The initial
+                // notification (changed == None) is always delivered so the
+                // subscriber receives the current value immediately on subscribe.
```
lukasmittag: I would remove this comment since it is not helping in the readability of the code :)

databroker/src/broker.rs
```diff
@@ -5060,4 +5062,67 @@ pub mod tests {
             }
         }
     }
+
+    /// Regression test for issue #186.
```
lukasmittag: maybe link the issue with a concrete web address?

lukasmittag (changes requested):

aki1770-del:
> @aki1770-del could you rebase the branch since else we can not merge. For your other PRs we need a strategy as well. Are there some comnflicts between them or an order we should merge them?

Hi @lukasmittag , thanks for the review. Rebasing #199 now — no conflicts with main. For the other PRs — they touch separate files so no ordering dependency, but if it helps for review, the suggested sequence would be: #196 (conversions.rs, 1-line) → #197 (JWT auth, self-contained) → #199 (broker.rs, this one). 
Changed as you requested
  1. Comment removed from broker.rs Some(changed) branch                                           
  2. Issue URL added: /// Regression test for issue #186 (<https://github.com/eclipse-kuksa/kuksa-databroker/issues/186>).

aki1770-del:
Hi @lukasmittag — the branch has been rebased onto current `main` (0 commits behind). All CI checks are green (integration tests on amd64/arm64/riscv64, unit tests, lint, license). Ready for re-review when you have a moment.

SebastianSchildt:
@mikehaller or @lukasmittag , "Technically" this looks fine now. Is this ready to merge from a logic perspective? Does it fix min_sample_interval handling as intended?

aki1770-del:
@SebastianSchildt to confirm the intent: the change enforces `min_sample_interval` as a floor between change-driven emissions. Prior behavior emitted on every value change regardless of `min_sample_interval`, which matched the VISS v2 spec for `onChange` semantics only if `min_sample_interval` were unset — so the original code silently ignored the configured floor. After this PR, if a subscriber has `min_sample_interval = X ms` and sets a change filter, emissions are throttled to at most one per X ms window while still respecting change semantics. Unit test in `broker.rs` exercises the throttling under a tight simulated clock. Let me know if a different semantic was intended; happy to adjust.

mikehaller:
Yes, the logic now handles `min_sample_interval` as intended.

SebastianSchildt (approved):
Thank you for this contribution.

lgtm 🍰

lukasmittag (approved):
LGTM now as well. Thanks!

PR #210, opened 2026-04-28
Fix capitalization in Docker files

SebastianSchildt:
Fixes FromAsCasing: 'as' and 'FROM' keywords' casing do not match

More info: https://docs.docker.com/go/dockerfile/rule/from-as-casing/


nayakned (approved):
lgtm

PR #211, opened 2026-04-28
Remove deprecated sdv.v1 API code

### LINKED ISSUE #202: Remove sdv 1 API code
Two years ago we deprecated sdv API and put it behind a commandline flag, due to it having some security issues. Since then kuksa.val.v1 and kuksa.val.v2 have been supported.

It is time to remove the sdv API code, as this would help us to get rid of some code, which is always good in terms of maintainability and stability.

SebastianSchildt:
This removes the SDV1 API that we have deprecated *2 years* ago.

This gets rid of > 13k lines of code that have not been maintained much anymore AND some dependencies, which is very good.

Main changes
 - remove all code related to SDV API
 - Remove mentions in documentation
 - Rewrote/refactored integration test to use Python SDK: This is the main "change" in this PR. The integration test was not so much for testing any API functionality, but has been used in CI to fire up containers for each architecture and trying to check if they are usable at all. It was based on "handcoded" SDV GRPC API. This has been changed to kuksa.val API via Python SDK

Not in scope of this PR:
 - Probably integraiton tests could be improved/extended quite a bit. But for this PR I kept the scope as it was before

Will fix #202


lukasmittag:
Are we sure we want to do this. I see a lot of people using the sdv.databroker.v1 from time to time? Shall we not continue supporting it? 

@mikehaller @BjoernAtBosch

mikehaller:
I think hardening databroker for production use and thus removing deprecated API is indeed more important than some of the features.

SebastianSchildt:
This is ready for merge/review now @lukasmittag  @mikehaller 

Plan is to release this as 0.7 as removing a deprecated API is more than just 0.6.2 , and this would give us the chance- IF hell freezes over and somebody really needs SDV1 or an urgent other fix on an SDV API enabled databroker, to still do 0.6.x later (I doubt it/hope it will not happen. But it is always good to have options :)

lukasmittag (commented):
is the removal of the whole gen_proto folder under integration__test intended?

SebastianSchildt:
@lukasmittag  Yes, my understanding is, that is not used by anything

The files were quite old, had no relation to  Rust, just Python, and I guess Py SDK generates such files by itself from some protos during build, or at least does not pull them from "integration test". The generated SDV files were in use by the old integration test, but of course within the scope of this PR are to be deleted, the kuksa.val generations looked like pure "decoration" to me, so I nuked the whole folder 

But anyway, this is why you review this, so feel free to  prove me wrong :-P

lukasmittag (approved):
Alright, sounds reasonable and I confirmed no one is using them as of now :)

PR #214, opened 2026-06-15
Update vergen-gix to version 10 to fix security warnings

SebastianSchildt:
Update vergen once again as we have 7 open security warnings about gix, that can only be upgraded if vergen is bumped as well

Minor GIX API changes accommodated for, no behavioral or output changes 

I used the chance to do a general cargo update to update all dependencies to the latest compatible patch levels as we need to do that before the coming release anyway. That lead to a breaking change in the http library (even though only the patch level had been updated), and I needed to add an "/" to the root path in URI contruction for behavior tests and CLI clients


lukasmittag (approved):
LGTM

PR #217, opened 2026-06-25
Upgrade dev container

SebastianSchildt:
I noticed the dev container was missing some components to build current databroker, and the rust version was too old to build/use the coverage tests.

So this is a dump of my local changes

- Moved to current version of the rust decvontainer
- add missing components (e.g. cmake used during prototyping comilation and llvm-cov package for cargo
- Removed some packages that seemed unneccesary

This "works for me", but maybe @mikehaller as our resident dev container fan, you could give it a try?


nayakned (approved):
lgtm

PR #218, opened 2026-06-29
Stronger grpc input data validation for kuksa.val.v2

SebastianSchildt:
A bad provider  could make a tonic thread panic when not sending a data_point in the publish_value request. This should be fixed now.

A similar fix makes sure a new provider registers a datapoint first before sending UpdateFilterResponse (I guess this would not happen by "accident", but if somebody did this, it would lead to a similar panic due to accessing non existent uuid in the debug log)


lukasmittag (approved):
LGTM

Issue #221, opened 2026-07-23
GetValue can panic on an empty provider response or return a stale/wrong signal value


s1wrucin:
[kuksa-databroker-getvalue-defect-20260723T044150Z.zip](https://github.com/user-attachments/files/30313181/kuksa-databroker-getvalue-defect-20260723T044150Z.zip)

I am working on an internal Software Defined Vehicle proof of concept at Spyrosoft Synergy that uses Eclipse KUKSA Databroker.
While testing the kuksa.val.v2 on-demand provider read path, I observed two related failure modes. I am reporting them together because both appear to occur at the same provider-response handling boundary, although they have different immediate effects.
I would appreciate confirmation whether my understanding of the provider contract is correct or whether the provider is expected to follow additional ordering constraints that are not obvious from the API.

Affected version
Image: ghcr.io/eclipse-kuksa/kuksa-databroker:0.7.0
Digest: sha256:9ace2824719ba7c05929f0af4827d9b25fa5ca7b2e0d58d0acff95885c649d6d
Commit: 70ea23a50a0efe62cf10a9be06faeed75845c5fa
Only version 0.7.0 was executed. The current main branch was inspected, but not runtime-tested.

Observed behavior
1. Empty provider response causes a panic
When an on-demand provider answers a GetProviderValueRequest with an empty entries map, the corresponding GetValue handling task panics at:
val.rs:238:59
Option::unwrap() on None
The client receives a cancelled HTTP/2 stream. In the collected runs, the Databroker process remained available after the task panic.
Reproduction rate: 10/10

2. A delayed response can be returned to a later request
When the response to request A arrives after request A has timed out, it can apparently be consumed by request B, which is waiting on the same provider connection.
In the strongest reproduction:
Client 1 requests FloatB.
The provider delays the response beyond the Databroker timeout.
Client 1 times out.
Client 2 requests FloatC.
The delayed FloatB response arrives during the wait for FloatC.
Client 2 receives the FloatB value with gRPC OK.
Both signals use the same protobuf value variant, so the incorrect result cannot be detected by checking only the gRPC status or value type.
Reproduction rate: 20/20

A separate scenario also confirmed that a provider response containing only a different signal ID can be accepted and returned with gRPC OK.
Reproduction rate: 20/20

Expected behavior
For an empty provider response, I would expect a defined gRPC error rather than a panic.
For delayed, mismatched, or multi-entry responses, I would expect the Databroker to either:
[...]

PR #222, opened 2026-07-30
WIP Gracefully handle empty provider response

SebastianSchildt:
Part of fixing #221 

This fixes the issue that an Empty provider Response triggers a panic.
(The other issue in #221 will be investigated and fixed separately)

Old logic for v2 API
 - A request for data is forwarded to a provider
 - If a provider provides data points they are forwarded to the caller
 - If a provider errors, the database is queried for the latest data

The issue:
 - If a provider returns "ok" but the response is empty, a panic can occur when databroker tries to unpack the none existing data (I think it is not clearly defined whether this is "correct" behavior of a provider but it might make sense for a provider when queried to say "I am here, but I have nt current data", in any way databroker should be robust against it)

The new logic:
  - We check for an empty response and handle it similarly as an error, i.e. trying the database feedback 

There also is a new unit test. The test has been fully AI generated... But I verified it triggers with the old logic, and passes now that the fix is in.

### DIFF
--- a/databroker/src/broker.rs
+++ b/databroker/src/broker.rs
@@ -2234,7 +2234,23 @@ impl AuthorizedAccess<'_, '_> {
                 //Step 4: collect responses from providers
                 match response {
                     Ok(value) => {
-                        entries_response.extend(value.entries);
+                        if value.entries.is_empty() {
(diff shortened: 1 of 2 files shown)


nayakned (approved):
Lgtm

PR #225, opened 2026-08-04
Fix processing of invalid/delayed provider responses

### LINKED ISSUE #221: GetValue can panic on an empty provider response or return a stale/wrong signal value

SebastianSchildt:
This should fix #221 

When a provider's response to request A arrived after request A had timed out, it could be consumed by request B, which was waiting on the same provider connection. 

Issues:
1. request_id was hardcoded to 0 for every GetProviderValueRequest, making it impossible to match a response to its originating request.
2. Shared broadcast channel — every get_signals_values_from_provider call subscribed to the same broadcast::Receiver. A response for any request was delivered to all concurrent waiters with no correlation.
3. No signal ID validation — get_value took the first map entry blindly (entries.values().next().unwrap()) without checking it matched the requested signal.

Fix
1. Provider struct — added next_request_id: u32 field, initialized to 1
2. get_signals_values_from_provider — added checks:
    - Send GetProviderValueRequest with a unique, incrementing request_id
    - Loop on the broadcast receiver with a 1s total deadline
    - Skips responses whose request_id doesn't match 
    - Filters entries to only include signal IDs present in the original request
    - Returns Err(()) if no matching entries remain (triggers existing DB fallback)

Also added some unit tests verifying correct processing of wrong replies

### DIFF
--- a/databroker/src/grpc/kuksa_val_v2/val.rs
+++ b/databroker/src/grpc/kuksa_val_v2/val.rs
@@ -3889,4 +3889,149 @@ mod tests {
             }
         }
     }
+
+    // ---- Provider response correlation (request_id / signal_id) tests ----
+
+    use crate::types::SignalId as TypesSignalId;
+
+    /// Helper: builds a Datapoint with a float value for use in broadcast response entries
+    fn make_float_datapoint(value: f32) -> proto::Datapoint {
+        proto::Datapoint {
+            timestamp: Some(std::time::SystemTime::now().into()),
+            value: Some(proto::Value {
+                typed_value: Some(proto::value::TypedValue::Float(value)),
+            }),
+        }
+    }
+
+    /// Creates a Provider directly for tests that control the broadcast receiver
+    fn make_provider_with_broadcast_rx(
+    ) -> (Provider, broadcast::Sender<proto::GetProviderValueResponse>) {
+        let (mpsc_tx, _mpsc_rx) =
+            mpsc::channel::<Result<OpenProviderStreamResponse, tonic::Status>>(10);
+        let (broadcast_tx, broadcast_rx) =
+            broadcast::channel::<proto::GetProviderValueResponse>(10);
+        let provider = Provider {
+            sender: mpsc_tx,
+            receiver: Some(BroadcastStream::new(broadcast_rx)),
+        };
+        (provider, broadcast_tx)
+    }
+
+    #[tokio::test]
+    async fn test_provider_correct_response_accepted() {
+        let our_signal = 1;
+        let (mut provider, broadcast_tx) = make_provider_with_broadcast_rx();
+
+        let mut entries = HashMap::new();
+        entries.insert(our_signal, make_float_datapoint(12.5));
+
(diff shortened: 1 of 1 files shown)


nayakned (approved):
lgtm

PR #229, opened 2026-08-26
Update dev version to 0.7.2-dev.0

SebastianSchildt:
After releasing 0.7.1 updating version tags in main

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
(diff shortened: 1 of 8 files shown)


lukasmittag (commented):
Cargo.lock update intended?

SebastianSchildt:
Yes, this is the now the same as in the release, according to release instructions https://github.com/eclipse-kuksa/kuksa-databroker/wiki/Release-Process

the scripts/prepare_release does a cargo update. In main we only do it "lazily" (e.g. when dependabot does it), or - after/before a release. But it should be updated from time to time, because the specific versions in .lock decide whether we are vulnerable or not...

lukasmittag (approved):
perfect than approved

